Issue #38, opened 2021-09-27
How to pass a numpy array into a Julia function?


ShuhuaGao:
Consider a dummy function inside my custom package EOS:
```julia
function dummy(x::AbstractVector)
    @show typeof(x)
end
```

From Python, I call it as follows
```python
jl.EOS.dummy(np.array([1.1, 2.3]))
```
and the following error occurs
```
TypeError: Julia: MethodError: no method matching dummy(::PythonCall.PyIterable{PythonCall.Py})
Closest candidates are:
  dummy(!Matched::AbstractVector{T} where T) at /home/shuhua/Work/opt/EOS/src/EOS.jl:26
```

According to [Conversion to Julia](https://cjdoris.github.io/PythonCall.jl/stable/conversion-to-julia/), I expect that a numpy array is converted to a Julia array automatically. Did I miss anything?

cjdoris:
Ah sorry, the package was recently rewritten and that functionality got overlooked - numpy arrays should automatically get converted to `PyArray` but that doesn't happen yet.

For now, on the Julia side, you can do `PyArray(x)` or `pyconvert(PyArray, x)` or `pyconvert(Array, x)` or `pyconvert(AbstractArray, x)` to convert `x` to a Julia array. From Python you can do `dummy = jl.eval("x->EOS.dummy(PyArray(x))")` then call `dummy(x)`.

ShuhuaGao:
Does it work on the Python side? I am using `juliacall` to pass a numpy array into a Julia function. The issue reported above still exists.

I first removed the old `juliacall` and then installed the latest master branch.
```
pip uninstall juliacall
pip install git+https://github.com/cjdoris/PythonCall.jl@master
```

cjdoris:
You need to do an editable install `pip install -e ...` to get the dev version of PythonCall, otherwise you get the latest release which doesn't have this fix yet.

ShuhuaGao:
Thanks. I tried
```
pip install -e git+https://github.com/cjdoris/PythonCall.jl@master#egg=juliacall
```
but failed to install it. 
```
Installing collected packages: juliacall
  Running setup.py develop for juliacall
    ERROR: Command errored out with exit status 1:
[...]

PR #39, opened 2021-09-29
Python setup

ShuhuaGao:
Add a *setup.py* such that even pip of an old version can work ([reference](https://stackoverflow.com/questions/46205648/can-pip-install-from-setup-cfg-as-if-installing-from-a-requirements-file)).  See issue #38 .

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,6 @@
+import site, sys, setuptools
+
+if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


ShuhuaGao:
> don't want too much developer-specific information

Makes sense. Done.

cjdoris:
Great, thanks!

PR #110, opened 2022-02-11
Reorganize init.py into functions with fewer side effects

jlapeyre:
This PR reorganizes most of the code in `init.py` into functions. I also reduced the use of globals/side-effects so that the required order of function calls is more clear.

A single, important, API change is also made. Rather than `import juliacall` doing all of the work of initializing Julia, you need to do this
```python
import juliacall
juliacall.default_init()
```

 [This](https://github.com/jlapeyre/PythonCall.jl/blob/1a5727d5ea5fa7c3c001978f33a45d082b281480/juliacall/init.py#L207-L226) is the initialization function
```python
def default_init():
    """
    Initialize Julia and its packages.
    After importing `juliacall`, call this method initialize the Julia
    dependencies. This includes first locating the julia executable and library,
    and if none is found, downloading a Julia distribution. Then the required
    Julia packages are installed.
    """
    # Are we developing PythonCall
    isdev = find_isdev()
    # Determine paths to mutable data maintained by PythonCall, and julia
    jlenv, jlprefix, meta_path = find_env_paths()
    # Determine whether or not to skip resolving julia/package versions
    skip = deps.can_skip_resolve(isdev, meta_path)
    # Find a compatible jula executable and libjulia
    libpath = find_julia_library(skip, jlprefix)
    # Initialze libjulia
    lib = init_libjulia(libpath)
    install_packages(lib, skip, isdev, jlenv, meta_path)
```

Part of the value is just in cleaning up the code: making data function-local, passing parameters rather than using global data, reducing the scope of a `try` block, etc.

The reason for delaying the initialization is that it gives a package author much more flexibility in designing a python package that depends on Julia (or even in a personal workflow) to meet particular needs.  So with this PR, one can immediately write a package-specific version of `default_init` and with a bit (or more) of code:

* pass a custom system image.
* Issue a prompt before downloading Julia.
* Use jill.py or juliaup to download Julia or search for the appropriate version.
* Things we don't know about, but someone else has in mind.

For example I am very interested in trying `juliacall` in [`julia_project`](https://github.com/jlapeyre/julia_project). This PR would make it much easier to integrate `juliacall` in such projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,15 @@ In this example, we use the Julia module PythonCall from a [Pluto](https://githu
 
 ## Example 2: Calling Julia from Python
 
+### Changes in jlapeyre branches !
+
+The examples have not been updated to reflect the following. In order to use the Python module JuliaCall (juliacall),
+you must do
+```python
+import juliacall
+juliacall.default_init()
+```
+
 In this example we use the Python module JuliaCall from an IPython notebook to train a simple neural network:
 - We generate some random training data using Python's Numpy.
 - We construct and train a neural network model using Julia's Flux.
--- a/juliacall/__init__.py
+++ b/juliacall/__init__.py
@@ -1,4 +1,7 @@
 __version__ = '0.5.1'
 CONFIG = dict(embedded=False)
 from .all import *
-from . import init
+
+# In application code, use `juliacall.default_init()`
+#from . import init
+from .init import default_init
--- a/juliacall/deps.py
+++ b/juliacall/deps.py
@@ -25,19 +25,17 @@ def julia_version_str(exe):
 
 META_VERSION = 1 # increment whenever the format changes
 
-def load_meta():
-    fn = CONFIG['meta']
-    if os.path.exists(fn):
-        with open(fn) as fp:
+def load_meta(meta_path):
+    if os.path.exists(meta_path):
+        with open(meta_path) as fp:
             meta = json.load(fp)
             if meta.get('meta_version') == META_VERSION:
                 return meta
 
-def save_meta(meta):
+def save_meta(meta_path, meta):
     assert isinstance(meta, dict)
     assert meta.get('meta_version') == META_VERSION
-    fn = CONFIG['meta']
-    with open(fn, 'w') as fp:
+    with open(meta_path, 'w') as fp:
         json.dump(meta, fp)
 
 ### RESOLVE
@@ -74,9 +72,9 @@ class PackageSpec:
         }
         return {k:v for (k,v) in ans.items() if v is not None}
 
-def can_skip_resolve():
+def can_skip_resolve(isdev_in, meta_path):
     # resolve if we haven't resolved before
-    deps = load_meta()
+    deps = load_meta(meta_path)
     if deps is None:
         return False
     # resolve whenever the version changes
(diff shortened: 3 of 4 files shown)


cjdoris:
Thanks! Yeah the init code is a mess, thanks for taking a look.

Note however I've also been refactoring the same code. Most of the code to find Julia and install dependencies is now in this package: https://github.com/cjdoris/pyjuliapkg. Take a look and let me know what you think.

My next task was to remove the corresponding parts of JuliaCall, which is almost all of it!

I won't however be changing JuliaCall to require the user to do `juliacall.init()` manually. For a start the user might not know they are using it, if they are using some other package which depends on JuliaCall. Also, it doesn't make much sense to allow packages to provide their own init function, because of two packages do this then they will clash. Instead JuliaCall should provide enough ways for other packages to declare what they need, and JuliaCall merges these requirements and satisfies them all when it initialises.

jlapeyre:
I played with `pyjulia` a bit. It looks great. I guess that separating this from `JuliaCall` will be the right move in the end. I like the support for juliaup. I'm curious about why you maintain your own downloading code rather than relying on jill.py. Maybe it doesn't meet your needs?

Regarding the manual `init`, you can instruct package authors to always put `juliacall.init()` after `import juliacall` unless they have a need that makes this impossible. If they do this, the result will be exactly what you have now. Of course you run the risk that someone might do something you don't want them to.

> Instead JuliaCall should provide enough ways for other packages to declare what they need

The problem with this approach is that people will have to wait until you implement a feature, and hope it fits their use case well enough, and hope you fix any bugs quickly, etc. They can't do any of this for themselves. For instance, I'd like to try the different approach to the interface between Julia and Python that juliacall offers. But, it only comes packaged with a lot of other stuff. And that other stuff does not meet my requirements unless I can have a bit of control during initialization. Of course many people can use juliacall without ever running into such a problem.

In any case, the project as a whole is really impressive. I'm keeping an eye on how it develops.

cjdoris:
Good point about jill.py. I looked into it before and it didn't provide enough control, e.g. you couldn't opt out of creating symlinks or editing the user's PATH. From a look at the readme it has been worked on since then so I'll look into using it. It would be nice to offload that part of the code.

I have now refactored JuliaCall on the main branch to use JuliaPkg. It's very short now. There's an env var to prevent the init function from being called - it's used by PythonCall to prevent JuliaCall from initialising when embedded, but feel free to use it yourself if you want to manually init the module.

I'm not going to require the user to call init. It's pretty unusual to require this - imagine if every package required a manual init, that would get pretty tedious.

jlapeyre:
> I played with pyjulia a bit. 

Oops, I meant I played with `pyjuliapkg`. 

There are a few other things about jil.py that I would change (well, if I could take the time). For example, it doesn't seem possible to install two patch versions of the same minor version at the same time. One thing I noticed is that as a library, jill.py takes about 60ms to load, whereas `pyjuliapkg`, `find_julia` and similar packages load in < 5ms. This is perhaps not a huge problem, but it increases the time significantly to use `find_julia` with `python -c` to return a path.  I suppose I could open issues for some of this.

> I have now refactored JuliaCall on the main branch to use JuliaPkg

This looks like a great approach.

> There's an env var to prevent the init function from being called

I was going to suggest this, but you've done it. This is perfect. It is enough to let me start exploring. I think putting the burden on the user who needs something out of the ordinary makes sense.

cjdoris:
I've now release PythonCall/JuliaCall with these changes. Closing this ticket now while you try it out.

Issue #117, opened 2022-02-21
Support selective import in @py macro `@py from <module> import ...`


hhaensel:
Currently partial imports, e.g. `@py from os.path import join as py_join` are not supported.

I propose to add an additional method to @py that supports this kind of statements.
I'll submit a PR, which you might consider if you like that feature.

BTW, just recently stumbled across `PythonCall` in the discourse announcements. I think it is a really cool package! 🚀

cjdoris:
As mentioned in #118 this functionality exists with syntax `@py import os.path: join as py_join`.

hhaensel:
Thanks for the feedback. I didn 't know that syntax and has just tried the simplest approach, my bad!
I like your version better than mine, it's closer to the Julia way of imports and therefore can be integrated into begin end clauses.

PR #118, opened 2022-02-21
add support for `@py from <module> import`

hhaensel:
This is the corresponding PR to #117 it allows for

```julia
julia> @py from os.path import join as py_joinpath, exists
(<py function exists at 0x00000000652A6B90>, <py function join at 0x00000000652A7400>)

julia> py_joinpath("test", "me")
Python str: 'test\\me'

julia> exists(".")
Python bool: True
```

### DIFF
--- a/src/py_macro.jl
+++ b/src/py_macro.jl
@@ -793,10 +793,26 @@ For example:
 - `x.foo` is translated to `pygetattr(x, "foo")`
 
 Compound statements such as `begin`, `if`, `while` and `for` are supported.
+Import statements are supported, e.g.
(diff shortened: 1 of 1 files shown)


cjdoris:
Thanks but this functionality already exists with a different syntax:
```julia
julia> @py import os.path: join as py_joinpath, exists

julia> py_joinpath("test", "me")
Python str: 'test\\me'

julia> exists(".")
Python bool: True
```

The `@py` macro doesn't intend to be as close as possible to Python syntax. Instead, it translates Julia syntax to the equivalent Python concept.

PR #123, opened 2022-02-24
crop whitespace around displayed figures

marius311:
This matches what PyPlot is [doing](https://github.com/JuliaPy/PyPlot.jl/blob/94366fcfd783e4467b9dae789c12a9762c87dcf3/src/PyPlot.jl#L100) too.


cjdoris:
Great idea, thanks!

PR #154, opened 2022-04-13
define 2-arg Base.hash

stevengj:
If you only define the 1-argument `hash(o::Py)`, then the 2-argument form `hash(o, h::UInt)` falls back to the generic `Base` method using `objectid(o)`, which is not what you want.

(Subtracting `- 3h` is what `Base.hash` typically does to mix hashes; I'm not sure if you want to do something fancier here?)


cjdoris:
Looks great, thanks.

PR #159, opened 2022-04-15
Accept optional argument in ArrayValue.__array__

moble:
I ran into a situation when passing an `SVector{4, Float64}` to a python function where `ArrayValue.__array__` was called with the optional `dtype` argument, which is [standard for numpy](https://numpy.org/devdocs/user/basics.dispatch.html), but caused an error here.  Adding this option to the call made the conversion work.

### DIFF
--- a/test/concrete.jl
+++ b/test/concrete.jl
@@ -1,4 +1,7 @@
 @testset "import" begin
+    @test pymoduleexists("sys")
+    @test pymoduleexists("os")
+    @test !pymoduleexists(randstring(32))
     sys = pyimport("sys")
     os = pyimport("os")
     @test pyeq(Bool, sys.__name__, "sys")
--- a/CondaPkg.toml
+++ b/CondaPkg.toml
@@ -1,3 +1,2 @@
 [deps]
 python = ">=3.5,<4"
-numpy = ""
--- a/CondaPkg.toml
+++ b/CondaPkg.toml
@@ -1,2 +1,3 @@
 [deps]
 python = ">=3.5,<4"
+numpy = ""
--- a/Project.toml
+++ b/Project.toml
@@ -25,7 +25,8 @@ julia = "1"
 
 [extras]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Aqua", "Test"]
+test = ["Aqua", "Random", "Test"]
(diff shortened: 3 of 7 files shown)


stevengj:
Ah looks like you ran into the same issue as is.  Presumably this PR fixes #160?

moble:
That's a pretty great github jinx!

I've added the test.  Testing this code path does require adding `numpy` as a dependency in `CondaPkg.toml`, but I think that's probably better anyway.

stevengj:
Maybe instead do:
```
pymoduleexists(mod::AbstractString) = pyconvert(Bool, pyimport("importlib").util.find_spec(mod) != Py(nothing))

if !pymoduleexists("numpy")
   PythonCall.C.CondaPkg.add("numpy")
end
```
so that it only installs numpy if the test is run?

moble:
I guess that makes sense,<sup>(*)</sup> though it'll evidently alter the `CondaPkg.toml` used to run the tests, which means that people making PRs and such will need to be careful not to commit that change...

That `pymoduleexists` function looks so handy that, IMHO, it should be exported.  I've added it to `src/concrete/import.jl` and exported it.  But this is just my suggestion; I'm open to differing opinions.

I might even suggest that a call like that to `CondaPkg` would form a nice basis for extending `pyimport` to act more like the new Julia `import` — prompting a user to add a package if it's not found, maybe with options for version and channel info.  The conversation by @icweaver and @sairus7 indicates there may be some interest in such functionality.

---
<sup>(*)</sup>  Except that it has to be `pyimport("importlib.util").find_spec(m)`.

stevengj:
> I might even suggest that a call like that to CondaPkg would form a nice basis for extending pyimport to act more like the new Julia import — prompting a user to add a package if it's not found, maybe with options for version and channel info. The conversation by @icweaver and @sairus7 indicates there may be some interest in such functionality.

PyCall has something like that with `pyimport_conda` (though without a prompt), but I think that probably should be a topic for another issue?

test/concrete.jl
```diff
@@ -1,4 +1,7 @@
 @testset "import" begin
+    @test pymoduleexists("sys")
+    @test pymoduleexists("os")
+    @test !pymoduleexists(randstring(32))
```
stevengj, line 4:
```suggestion
    @test !pymoduleexists("rDJb5uy3eTorFw4c7UbcGOC59FjVGcnB") # randstring(32)
```
No reason to choose a different random string for each test run.

stevengj: (Then you can also remove the `Random` dependency.)

cjdoris:
Thanks for the PR! Could you revert all the changes except those to array.jl please? I'm happy to discuss pymoduleexists in another ticket, though my gut is to just let the user call find_spec themselves. I don't mind this not being tested right now, this clearly works, and there is a lot of test writing to be done anyway.

cjdoris:
I've just pushed the changes to `__array__()`. Happy to discuss the other changes on a separate ticket.

PR #167, opened 2022-05-08
fix relative JULIA_PROJECT

marius311:
Previously this would fail if e.g. `JULIA_PROJECT=.` because of the chdir just  below.


cjdoris:
I believe this is not relevant given the other PR is not being merged. Anyway `project()` should be absolute already, it's a bug if not.

PR #169, opened 2022-05-09
Fix minor bug in `Utils.is{c,f}contiguous`

dingraha:
PythonCall.jl is awesome!

I was curious if NumPy arrays passed to Julia were C-contiguous or Fortran-contiguous, and ran into a minor bug in `Utils.isccontiguous` and `Utils.isfcontiguous`. Fixed it, added some tests.


cjdoris:
Thank you, but those are internal functions and were not being used anywhere, so I have just deleted them.

PR #186, opened 2022-06-17
fix: Ensure julia lib is always loaded as global

gnuille:
From issue https://github.com/cjdoris/PythonCall.jl/issues/175 some users have noticed that there is an issue with running the Juliacall python module on Apple.

Investigating the issue was found that the julia library load was causing the error.

From my report and thanks to the Julia community they pointed out that the library should be loaded in RTLD_GLOBAL mode (https://github.com/JuliaLang/julia/issues/45709)

### DIFF
--- a/pysrc/juliacall/__init__.py
+++ b/pysrc/juliacall/__init__.py
@@ -118,7 +118,7 @@ def init():
     try:
         # Open the library
         os.chdir(os.path.dirname(libpath))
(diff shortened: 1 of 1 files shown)


MilesCranmer:
Confirmed that this fixes juliacall on my M1 mac! Amazing work @gnuille.

cjdoris:
This doesn't appear to break anything on supported platforms. Great that it was such a simple fix to get it working on M1. Thanks!

PR #192, opened 2022-06-27
pass settings from -Xjuliacall_ to julia

### LINKED ISSUE #184: Configure julia options
Hi,
with pyjulia it's possible to configure some julia options (see https://github.com/JuliaPy/pyjulia/blob/master/src/julia/options.py#L79)
Is it possible with PythonCall?
If not, I'm willing to create a PR if somebody could give me any hint on how to integrate it

related to https://github.com/cjdoris/PythonCall.jl/issues/73

dpinol:
added support for --threads, --procs, --min-optlevel

Fixes #184

### DIFF
--- a/pysrc/juliacall/__init__.py
+++ b/pysrc/juliacall/__init__.py
@@ -3,6 +3,9 @@
 
 __version__ = '0.9.1'
 
+from ctypes import pointer, c_int, POINTER, c_char_p
+from typing import Any
+
 _newmodule = None
 
 def newmodule(name):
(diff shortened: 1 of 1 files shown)


cjdoris:
Great PR, thank you! Most of the options seem to work great, but some require special handling:
- `opt_bindir` translates to `--home` not `--bindir`. Could rename it `opt_home`?
- `opt_bindir` and `opt_sysimage` are parsed by `jl_parse_opts` but have no further effect. I think the call to `jl_init__threading` needs to be replaced by `jl_init_with_image__threading`.
- `opt_procs` also has no effect, it's handled separately in Base here (but is only called later in the loading process than `jl_init`): https://github.com/JuliaLang/julia/blob/762561c3057b4bd541d183ca1d1b70b9945bff22/base/client.jl#L245-L254

The first one should be fixed in this PR. I'm happy to leave the others for separate PRs (and not document them yet) - it's up to you.

Speaking of `bindir`, there is the Julia env var `JULIA_BINDIR` and the command line option `--home`, hence the naming discrepancy in the first bullet point. It seems like JuliaCall should have similar behaviour (`PYTHON_JULIACALL_BINDIR` and `-X juliacall_home`) to be consistent with Julia. Opinions?

cjdoris:
I've just pushed a change to the main branch which, when merged with this PR, means that `opt_procs` works.

thautwarm:
> 

After merging this, I could make a PR for supporting `opt_bindir` and `opt_sysimage`.

dpinol:
thank you guys!

cjdoris:
I've merged this PR and made a few tweaks, to get the bindir/home option working. I've also changed the `-X` options to use `-`-separated words instead of `_`. And they are documented here: https://cjdoris.github.io/PythonCall.jl/dev/juliacall/#julia-config

A PR to support bindir and sysimage would be great.

PR #243, opened 2022-11-08
libstdcxx version-bounds

### LINKED ISSUE #237: libstdcxx-ng versioning
Thanks for the excellent package! I recently ran into a situation where I tried to install a Python package using CondaPkg.jl that depended upon libstdcxx-ng v12 and was unable to install it due to incompatible dependencies. This resulted in some error about GLIBCXX_3.4.30 not being found. I think I could in principle resolve this by building Julia from source with a more recent version of libstdc++, but because of the lines below I don't think it would work even then. Would it be a reasonable idea to actually run the `strings` command and find the highest GLIBCXX version string in PythonCall?

https://github.com/cjdoris/PythonCall.jl/blob/77ee463d149b418ed1e6b34036f50172429ae46e/src/cpython/context.jl#L67-L80

Also, the link in the comment on L70 doesn't contain GLIBCXX_3.4.30, however I do see it mentioned here: https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html

samtkaplan:
Hello,

We have run into a problem when using PythonCall with newer versions of TensorFlow.  In particular, TensorFlow requires newer versions of libstdc++ than what is bundled with Julia.  To partially solve this problem, we replace the bundled version of libstdc++ with the system (Ubuntu 22.04) version of libstdc++.

When using PythonCall and CondaPkg, PythonCall injects a version bound for libstdc++ into the conda operations.  The version bound is determined by the version of libstdc++ bundled in Julia.  In this PR, we find the bound at runtime, and allow the user to specify the version bound via an environment variable: ~`JULIA_CONDAPKG_LIBSTDCXX_VERSION_BOUND`~`JULIA_PYTHONCALL_LIBSTDCXX_VERSION_BOUND`.  For example, set the version bound via:
```julia
ENV["JULIA_PYTHONCALL_LIBSTDCXX_VERSION_BOUND"] = ">=3.4,<=12"
```

Many thanks!

Fixes #237

### DIFF
--- a/src/cpython/context.jl
+++ b/src/cpython/context.jl
@@ -112,7 +112,7 @@ function init_context()
                     if cxx_version !== nothing
                         CondaPkg.add("libstdcxx-ng", version=cxx_version, channel="conda-forge", temp=true, file=joinpath(@__DIR__, "..", "..", "CondaPkg.toml"), resolve=false)
                     end
-                    # if cxx_version is nothing, then we assume that Julia does not link against libstdcxx-ng, and so we do not
+                    # if cxx_version is nothing, then we assume that Julia does not link against any version ob libstdc++ known by Julia, and so we do not
                     # enforce a version bound.
                 end
                 # By default, we use Python installed by CondaPkg.
--- a/src/cpython/context.jl
+++ b/src/cpython/context.jl
@@ -67,7 +67,7 @@ function get_libstdcxx_version_bound()
     # the highest GCC version we know about, which should be a fairly safe choice.
     max_version = get(vers_mapping, loaded_libstdcxx_version.patch, vers_mapping[maximum(keys(vers_mapping))])
     cxx_version = ">=3.4,<=$(max_version.major).$(max_version.minor)"
-    get(ENV, "JULIA_CONDAPKG_LIBSTDCXX_VERSION_BOUND", cxx_version)
+    get(ENV, "JULIA_PYTHONCALL_LIBSTDCXX_VERSION_BOUND", cxx_version)
 end
 
 function init_context()
--- a/src/cpython/context.jl
+++ b/src/cpython/context.jl
@@ -44,14 +44,29 @@ end
 #   https://gcc.gnu.org/onlinedocs/gcc-12.1.0/libstdc++/manual/manual/abi.html
 # for the highest GCC version compatible with the highest GLIBCXX version.
 function get_libstdcxx_version_bound()
-    if Base.VERSION <= v"1.6.2"
-        # GLIBCXX_3.4.26
-        cxx_version = ">=3.4,<9.2"
-    else
-        # GLIBCXX_3.4.29
-        # checked up to v1.8.0
-        cxx_version = ">=3.4,<11.4"
-    end
+    # This list comes from: https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html
+    # Start with GCC 4.8, as it's extremely difficult to build Julia with anything older
+    vers_mapping = Dict(
+        18 => v"4.8.0",
+        19 => v"4.8.3",
+        21 => v"5.1.0",
+        22 => v"6.1.0",
+        23 => v"7.1.0",
+        24 => v"7.2.0",
+        25 => v"8.1.0",
+        26 => v"9.1.0",
+        27 => v"9.2.0",
+        28 => v"9.3.0",
+        29 => v"11.1.0",
+        30 => v"12.1.0",
+        31 => v"13.1.0",
+    )
+    # Get the libstdcxx version that is currently loaded in this Julia process
+    loaded_libstdcxx_version = Base.BinaryPlatforms.detect_libstdcxx_version()
+    # Map it through to get a GCC version; if the version is unknown, we simply return
+    # the highest GCC version we know about, which should be a fairly safe choice.
+    max_version = get(vers_mapping, loaded_libstdcxx_version.patch, vers_mapping[maximum(keys(vers_mapping))])
+    cxx_version = ">=3.4,<=$(max_version.major).$(max_version.minor)"
     get(ENV, "JULIA_CONDAPKG_LIBSTDCXX_VERSION_BOUND", cxx_version)
 end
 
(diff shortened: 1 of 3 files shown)


src/cpython/context.jl
```diff
@@ -30,6 +30,31 @@ function _atpyexit()
...
+# for the highest GCC version compatible with the highest GLIBCXX version.
+function get_libstdcxx_version_bound()
+    if Base.VERSION <= v"1.6.2"
+        # GLIBCXX_3.4.26
+        cxx_version = ">=3.4,<9.2"
+    else
+        # GLIBCXX_3.4.29
+        # checked up to v1.8.0
+        cxx_version = ">=3.4,<11.4"
+    end
```
staticfloat, line 54: We can actually dynamically probe for the currently-loaded `libstdc++`'s capabilities, by using [code such as `Base.BinaryPlatforms.detect_libstdcxx_version()`](https://github.com/JuliaLang/julia/blob/317211a38947d1d72a7471186490fa9df4944201/base/binaryplatforms.jl#L859-L886).  That will return a `VersionNumber` such as `v3.4.29`.

We can then translate this to a GCC release (as Conda appears to desire) by using a dictionary lookup on the `.patch` member:

```suggestion
    # This list comes from: https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html
    # Start with GCC 4.8, as it's extremely difficult to build Julia with anything older
    vers_mapping = Dict(
        18 => v"4.8.0",
        19 => v"4.8.3",
        21 => v"5.1.0",
        22 => v"6.1.0",
        23 => v"7.1.0",
        24 => v"7.2.0",
        25 => v"8.1.0",
        26 => v"9.1.0",
        27 => v"9.2.0",
        28 => v"9.3.0",
        29 => v"11.1.0",
        30 => v"12.1.0",
        31 => v"13.1.0",
    )
    # Get the libstdcxx version that is currently loaded in this Julia process
    loaded_libstdcxx_version = Base.BinaryPlatforms.detect_libstdcxx_version()
    # Map it through to get a GCC version; if the version is unknown, we simply return
    # the highest GCC version we know about, which should be a fairly safe choice.
    max_version = get(vers_mapping, loaded_libstdcxx_version.patch, vers_mapping[maximum(keys(vers_mapping))])
    cxx_version = ">=3.4,<=$(max_version.major).$(max_version.minor)"
```

src/cpython/context.jl
```diff
@@ -30,6 +30,46 @@ function _atpyexit()
...
+        22 => v"6.1.0",
+        23 => v"7.1.0",
+        24 => v"7.2.0",
+        25 => v"8.1.0",
+        26 => v"9.1.0",
+        27 => v"9.2.0",
+        28 => v"9.3.0",
+        29 => v"11.1.0",
+        30 => v"12.1.0",
+        31 => v"13.1.0",
```
cjdoris, line 62: Looking at https://gcc.gnu.org/develop.html#timeline there are more versions of GCC than in that other linked page. If I understand right, I think it means that we could bump the mappings to `28 => 9.5.0`, `29 => 11.3.0`, `30 => 12.2.0`.

src/cpython/context.jl
```diff
@@ -30,6 +30,46 @@ function _atpyexit()
...
+    )
+    # Get the libstdcxx version that is currently loaded in this Julia process
+    loaded_libstdcxx_version = Base.BinaryPlatforms.detect_libstdcxx_version()
```
cjdoris, line 65: We need to gracefully handle the possibility that this returns `nothing` or a version other than `3.4.*`.

src/cpython/context.jl
```diff
@@ -30,6 +30,46 @@ function _atpyexit()
...
+    # Map it through to get a GCC version; if the version is unknown, we simply return
+    # the highest GCC version we know about, which should be a fairly safe choice.
+    max_version = get(vers_mapping, loaded_libstdcxx_version.patch, vers_mapping[maximum(keys(vers_mapping))])
+    cxx_version = ">=3.4,<=$(max_version.major).$(max_version.minor)"
+    get(ENV, "JULIA_CONDAPKG_LIBSTDCXX_VERSION_BOUND", cxx_version)
```
cjdoris, line 70: Since this is an option to PythonCall, please rename the env var to `JULIA_PYTHONCALL_...`.

cjdoris:
This is really cool, I didn't know you could query the libstdc++ version like this!

src/cpython/context.jl
samtkaplan: fixed. thanks.

src/cpython/context.jl
samtkaplan: I went ahead and applied you suggestion.  If I understand correctly, it seems that your suggestion is safe since libstdc++ should be backwards compatible.

cjdoris: The ABI Policy page says "If a particular release is not listed, it has the same version labels as the preceding release." I take that to mean that all releases >=9.3.0 and <10.1.0 have the same version labels.

src/cpython/context.jl
samtkaplan: I added some code in the last commit (https://github.com/cjdoris/PythonCall.jl/pull/243/commits/e7c8dac5642b28fa0c312ad0ff4fae9ce6658133) that attempts to address this, but I'm not sure what the best solution is.  In particular, if Julia is not able to detect the version, then we try to get the version bound from the user environment variable, and if the user has not set a variable, then do not inject a version bound.  Hopefully the last case will be rare.

cjdoris:
Great, thank you!

cjdoris:
@samtkaplan FYI I have moved this functionality into CondaPkg, so that PythonCall doesn't need to hackily add a new dependency at run-time - it instead depends on libstdcxx_ng at a special version "<=julia" which is now replaced by CondaPkg with an appropriate version (or removed entirely if not on Linux). The end result is the same.

https://github.com/cjdoris/CondaPkg.jl/commit/cd53c00170a60e70b3225b30f9d13e0815e62403

https://github.com/cjdoris/PythonCall.jl/commit/66355c568c96fb93531b6756467b65c3c3556cbd

PR #249, opened 2022-11-26
add precompilation statements, using `SnoopPrecompile`

t-bltg:
Has something like this been considered to reduce latency ?
Obviously, I do not know the internals of `PythonCall`, and this would need more extensive testing, but I think you get the idea.

Xref https://github.com/cjdoris/PythonCall.jl/issues/45, for https://github.com/JuliaPlots/Plots.jl/pull/4542.

**pr**
```julia
julia> @time using PythonCall
  7.115765 seconds (3.18 M allocations: 213.703 MiB, 2.23% gc time, 86.78% compilation time: 1% of which was recompilation)
```

**master**
```julia
julia> @time using PythonCall
  9.773912 seconds (7.86 M allocations: 454.046 MiB, 4.12% gc time, 89.42% compilation time: 73% of which was recompilation)
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,6 +13,7 @@ Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
+SnoopPrecompile = "66db9d55-30c0-4569-8b51-7e840670fc0c"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 UnsafePointers = "e17b2a0c-0bdf-430a-bd0c-3a23cae4ff39"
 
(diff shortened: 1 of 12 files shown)


cjdoris:
Thanks for the PR! Nice idea to precompile all those init functions.

I've tried SnoopPrecompile in the past and not had much luck. Indeed on my computer, this PR makes importing PythonCall 0.4 seconds slower - 3.9 seconds on main branch vs 4.3 seconds here. Admittedly the 2.6 second gain on your computer is probably worth a small regression on mine. It would be great (but hard, probably) to understand the difference.

I also need to consider the effect of running all those init functions at precompile time. It's not obvious it won't mess something up - the fact that you needed to comment a few out due to segfaults backs this up. Will need to dig deeper.

I'm also hoping that upcoming native code cacheing in Julia will just magically take care of this sort of thing. 🤞

t-bltg:
Obviously this approach is quite unsafe in terms of pointers, yes.

I've observed these segfaults in https://github.com/JuliaPlots/Plots.jl/pull/4566 when trying to precompile the `pyplot` or `pythonplot` backends to improve ttfp. A part of the initialization still has to be done at runtime since you still have to initialize the `python` interpreter somehow.

Thanks for the comments here, my initial goal was to reach parity with `pyplot`, and this was achieved after the discussions in https://github.com/cjdoris/PythonCall.jl/issues/45. We'd have to be more robust in order to use a `SnoopPrecompile` approach. I think the native code caching will speed things up, but I still think it will segfault because of the pointer nullification.

Anyway, let's close this for now: it was a first naive approach.

PR #288, opened 2023-03-31
Add link in faq.md to instructions to run Julia within Blender

cdsousa:


frankier:
I guess this should be opened on JuliaCall rather than PythonCall?

cjdoris:
Great thank you!

PR #332, opened 2023-06-28
print caught exceptions when (de)serializing

glennmoy:
The `caught` exception can useful for tracking down sources of bugs, as was the case for myself earlier today.

### DIFF
--- a/src/cpython/jlwrap.jl
+++ b/src/cpython/jlwrap.jl
@@ -205,10 +205,13 @@ function _pyjl_serialize(self::PyPtr, ::PyPtr)
         b = take!(io)
         return PyBytes_FromStringAndSize(pointer(b), sizeof(b))
     catch e
-        PyErr_SetString(
-            POINTERS.PyExc_Exception,
-            "error serializing this value. Caught exception $(sprint(showerror, e, catch_backtrace()))"
-        )
+        PyErr_SetString(POINTERS.PyExc_Exception, "error serializing this value")
+        # wrap sprint in another try-catch block to prevent this function from throwing
+        try
+            @debug "Caught exception $(sprint(showerror, e, catch_backtrace()))"
+        catch e2
+            @debug "Error printing exception: $e2"
+        end
         return PyNULL
     end
 end
@@ -223,10 +226,13 @@ function _pyjl_deserialize(t::PyPtr, v::PyPtr)
         x = deserialize(io)
         return PyJuliaValue_New(t, x)
     catch e
-        PyErr_SetString(
-            POINTERS.PyExc_Exception,
-            "error deserializing this value. Caught exception $(sprint(showerror, e, catch_backtrace()))"
-        )
+        PyErr_SetString(POINTERS.PyExc_Exception, "error deserializing this value")
+        # wrap sprint in another try-catch block to prevent this function from throwing
+        try
+            @debug "Caught exception $(sprint(showerror, e, catch_backtrace()))"
+        catch e2
+            @debug "Error printing exception: $e2"
+        end
         return PyNULL
     end
 end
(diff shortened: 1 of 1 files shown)


src/cpython/jlwrap.jl
```diff
@@ -220,7 +223,10 @@ function _pyjl_deserialize(t::PyPtr, v::PyPtr)
         x = deserialize(io)
         return PyJuliaValue_New(t, x)
     catch e
-        PyErr_SetString(POINTERS.PyExc_Exception, "error deserializing this value")
+        PyErr_SetString(
+            POINTERS.PyExc_Exception,
+            "error deserializing this value. Caught exception $e"
+        )
```
haberdashPI, line 229: If you're printing out the exception, why not also include the trace? e.g.
```suggestion
        PyErr_SetString(
            POINTERS.PyExc_Exception,
            "error deserializing this value. Caught exception $(sprint(showerror, e. catch_backtrace()))"
        )
```

glennmoy: good idea!

src/cpython/jlwrap.jl
```diff
@@ -205,7 +205,10 @@ function _pyjl_serialize(self::PyPtr, ::PyPtr)
         b = take!(io)
         return PyBytes_FromStringAndSize(pointer(b), sizeof(b))
     catch e
-        PyErr_SetString(POINTERS.PyExc_Exception, "error serializing this value")
+        PyErr_SetString(
+            POINTERS.PyExc_Exception,
+            "error serializing this value. Caught exception $e"
```
glennmoy, line 210:
```suggestion
            "error serializing this value. Caught exception $(sprint(showerror, e. catch_backtrace()))"
```

src/cpython/jlwrap.jl
```diff
@@ -205,7 +205,10 @@ function _pyjl_serialize(self::PyPtr, ::PyPtr)
         b = take!(io)
         return PyBytes_FromStringAndSize(pointer(b), sizeof(b))
     catch e
-        PyErr_SetString(POINTERS.PyExc_Exception, "error serializing this value")
+        PyErr_SetString(
+            POINTERS.PyExc_Exception,
+            "error serializing this value. Caught exception $(sprint(showerror, e. catch_backtrace()))"
```
glennmoy, line 210:
```suggestion
            "error serializing this value. Caught exception $(sprint(showerror, e, catch_backtrace()))"
```

src/cpython/jlwrap.jl
```diff
@@ -220,7 +223,10 @@ function _pyjl_deserialize(t::PyPtr, v::PyPtr)
         x = deserialize(io)
         return PyJuliaValue_New(t, x)
     catch e
-        PyErr_SetString(POINTERS.PyExc_Exception, "error deserializing this value")
+        PyErr_SetString(
+            POINTERS.PyExc_Exception,
+            "error deserializing this value. Caught exception $(sprint(showerror, e. catch_backtrace()))"
```
glennmoy, line 228:
```suggestion
            "error deserializing this value. Caught exception $(sprint(showerror, e, catch_backtrace()))"
```

cjdoris:
I wonder if it's better to print a `@debug` log message instead.

Either way, please wrap this all in another try-catch block in case the `sprint` call fails for whatever reason. It's essential that this code doesn't throw.

glennmoy:
> I wonder if it's better to print a `@debug` log message instead.
> 
> Either way, please wrap this all in another try-catch block in case the `sprint` call fails for whatever reason. It's essential that this code doesn't throw.

sure - let me know if you think it's a good idea to print a second error message in the `catch` block... I figured it should say it tried to do display the original error without getting too self-referential.

cjdoris:
Looks good, thanks!

Issue #330, opened 2023-06-28
Support for Multi-Threading with PythonCall?


nic-barbara:
Are there any plans to support multi-threading with `PythonCall.jl` in the near future? Almost everything about this package is superior to `PyCall.jl` except for thread safety.

Specifically, it would be great (and in my case - is essential) to be able to call python functions on many threads in parallel. This is useful when a core routine is written in python and needs to be called many time independently. 

For example, some environments in [`ReinforcementLearning.jl`](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl) rely on the python bindings to [MuJoCo](https://mujoco.readthedocs.io/en/stable/overview.html). Evaluating these environments efficiently requires multi-threading (eg: [this implementation](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/72d67668f52ea45e90ee79b0a3a33055cce8cef7/src/ReinforcementLearningZoo/src/algorithms/policy_gradient/multi_thread_env.jl) with `PyCall.jl`), which is currently impossible in `PythonCall`.

Is this something you would consider supporting in the near future?

cjdoris:
Because of the GIL, Python itself has poor support for multithreading, because only one thread can hold the GIL at a time, and this is required to do anything Pythonic. Hence multithreading is only useful in Python if you are doing a lot of work which can release the GIL (such as IO or long-running numerical routines written in C).

In general, if you want to parallelise Python it's generally better to use multiprocessing instead of multithreading. So there are no plans for PythonCall to support it.

Issue #298, opened 2023-04-25
Bus error with multithreading linear algebra


aditya-sengupta:
**Affects:** JuliaCall

**Describe the bug**
I'm running into a bus error in trying to port and speed up a linear algebra routine from Python to Julia.

In an environment with numpy and juliacall, the following code produces the error:
```
# jl_seg.py
import numpy as np
from juliacall import Main as jl

np.random.seed(100)

jl.seval("using Base.Threads")

jl.seval("""
function tri_solve_vec_col_b(N, a, b, c, r, g, u)
    beta = b[1]
    u[1] = r[1] / beta

    @inbounds @fastmath for j in 2:N
        g[j] = c[j-1] / beta
        beta = b[j] - a[j] * g[j]
        u[j] = (r[j] - a[j] * u[j-1]) / beta
    end
    @inbounds @fastmath for k in N-1:-1:1
        u[k] -= g[k+1] * u[k+1]
    end
end
""")

jl.seval("coln(x,i) = view(x,:,i)")

tri_solve_vec_b = jl.seval("""
    function tri_solve_vec_b(a, b, c, r, g, u)
    N = size(a, 1)
    Threads.@threads for i in 1:N
        tri_solve_vec_col_b(N, coln(a,i), coln(b,i), coln(c,i), coln(r,i), coln(g,i), coln(u,i))
    end
end
""")

mats = [np.random.random((10,10)) for _ in range(6)]
tri_solve_vec_b(*mats)
```
When run with `python jl_seg.py`, this produces the error message
`[1]    73474 bus error  python jl_seg.py`
The number preceding the "bus error" changes each time.

**Your system**
Please provide detailed information about your system:
- MacOS Ventura 13.2.1 on a 2021 MacBook Pro with Apple M1 Pro chip
- Julia 1.8.5, Python 3.11.2, JuliaCall 0.9.12
```python -m pip list                                                                                                                                                                                  
Package          Version
---------------- -------
juliacall        0.9.12
juliapkg         0.1.10
numpy            1.24.3
pip              23.1.1
semantic-version 2.10.0
setuptools       65.6.3
```

```
>>> juliapkg.status()
JuliaPkg Status
/Users/adityasengupta/projects/misc/venv/julia_env/pyjuliapkg/juliapkg.json (empty project)
Julia 1.8.5 @ /Users/adityasengupta/.julia/juliaup/julia-1.8.5+0.aarch64.apple.darwin14/bin/julia
```

[...]

Issue #219, opened 2022-09-09
segmentation fault with multi-threading


lmiq:
I have this MWE, where I get segmentation faults (frequently, but not deterministically), when trying to run some script that uses multi-threading on the Julia side. 

I have used before launching `ipython3`:

```bash
export JULIA_NUM_THREADS=4
```

(my computer has 4 cores - 8 threads).

The MWE is:

```python
Python 3.10.4 (main, Jun 29 2022, 12:14:53) [GCC 11.2.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 7.31.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from juliacall import Main as jl

In [2]: import numpy as np

In [3]: jl.seval("""
   ...: function test(x)
   ...:     partial = zeros(Threads.nthreads())
   ...:     Threads.@threads for i in 1:Threads.nthreads()
   ...:         for j in i:Threads.nthreads():length(x)
   ...:             partial[i] += x[j]
   ...:         end
   ...:     end
   ...:     return sum(partial)
   ...: end
   ...: """)
Out[3]: test (generic function with 1 method)

In [4]: x = np.random.random((10_000,))

In [5]: %timeit jl.test(x)
72.2 µs ± 35.9 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [6]: %timeit jl.test(x)
Segmentation fault (core dumped)

```

Here I have emulated the error using the `%timeit`  macro from `ipython`, but my actual error I get after some runs of a function of my package:

```python
In [1]: from juliacall import Main as jl

In [2]: jl.seval("using CellListMap")

In [3]: import numpy as np

In [5]: x = np.random.random((10_000,3))

In [6]: nb = jl.neighborlist(x.transpose(), 0.05)

In [7]: nb = jl.neighborlist(x.transpose(), 0.05)

In [8]: nb = jl.neighborlist(x.transpose(), 0.05)

In [9]: nb = jl.neighborlist(x.transpose(), 0.05)

In [10]: nb = jl.neighborlist(x.transpose(), 0.05)

In [11]: nb = jl.neighborlist(x.transpose(), 0.05)

In [12]: nb = jl.neighborlist(x.transpose(), 0.05)

In [13]: nb = jl.neighborlist(x.transpose(), 0.05)

In [14]: nb = jl.neighborlist(x.transpose(), 0.05)

[...]

PR #333, opened 2023-06-28
allow user to configure opt_handle_signals

brian-dellabetta:
Hi @cjdoris , this PR is to allow a user to configure what is passed to `julia --handle-signals=<yes,no>`. I have just updated it to be in line with all the other CONFIG params, defaulting to `no` to retain backwards compatibility. 

Following discussion thread starting [here](https://github.com/cjdoris/PythonCall.jl/issues/219#issuecomment-1605087024), I have found that this is much more robust and clean way to allow for multi-threaded julia code than [my previous workaround](https://github.com/cjdoris/PythonCall.jl/issues/298#issuecomment-1589784428).

Related to #219 
Related to #298 
Related to #330

### DIFF
--- a/pysrc/juliacall/__init__.py
+++ b/pysrc/juliacall/__init__.py
@@ -129,7 +129,7 @@ def init():
     CONFIG['opt_sysimage'] = sysimg = path_option('sysimage', check_exists=True)[0]
     CONFIG['opt_threads'] = int_option('threads', accept_auto=True)[0]
     CONFIG['opt_warn_overwrite'] = choice('warn_overwrite', ['yes', 'no'])[0]
-    CONFIG["opt_handle_signals"] = choice('handle_signals', ['yes', 'no'], default='no')[0]
+    CONFIG['opt_handle_signals'] = choice('handle_signals', ['yes', 'no'], default='no')[0]
 
     # Stop if we already initialised
     if CONFIG['inited']:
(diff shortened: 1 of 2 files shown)


cjdoris:
Thank you. Could you also update the FAQ.md please with better advice for multithreading?

pysrc/juliacall/__init__.py
```diff
@@ -129,7 +129,7 @@ def args_from_config():
     CONFIG['opt_sysimage'] = sysimg = path_option('sysimage', check_exists=True)[0]
     CONFIG['opt_threads'] = int_option('threads', accept_auto=True)[0]
     CONFIG['opt_warn_overwrite'] = choice('warn_overwrite', ['yes', 'no'])[0]
-    CONFIG['opt_handle_signals'] = 'no'
+    CONFIG['opt_handle_signals'] = choice('handle_signals', ['yes', 'no'], default='no')[0]
```
giordano, line 132: Can we default to `yes`?  Defaulting to a behaviour which only causes crashes for backward compatibility doesn't seem very useful.

brian-dellabetta:
@cjdoris also replied to corresponding issue [here](https://github.com/cjdoris/PythonCall.jl/issues/219#issuecomment-1623924612). I think users would want signal handling on by default, should we default to `yes`? It's a behavioral change from previous versions, but almost always wanted

pysrc/juliacall/__init__.py
brian-dellabetta: > Can we default to `yes`? Defaulting to a behaviour which only causes crashes for backward compatibility doesn't seem very useful.

I agree, that's probably better to default to `yes`, but I left off for now as it is a behavioral change. Just posted a comment on this thread too, let's see what the author thinks

cjdoris:
Thanks! Please open a separate issue/PR to change the default - AFAIR making it yes by default has undesirable side-effects on Python.

Issue #293, opened 2023-04-13
`DataFrame(::PyPandasDataFrame)` converts date & datetime to bytes


tomdstone:
When doing some work involving dataframes in python via PythonCall, it seems like `DataFrame(PyTable(p))` where `p` is a pandas data table converts the date and datetime columns into byte vectors.  Is this issue related to the issue #265 with milliseconds vs microseconds, or due to a missing part of the `DataFrame(::PyPandasDataFrame)` implementation?

Here are a few minimal examples, in a conda environment with pandas.

```
using PythonCall
using Dates
using DataFrames

a = DataFrame(x = [now()])  # julia dataframe
b = pytable(a)              # pandas dataframe
c = PyTable(b)              # PyPandasDataFrame
d = DataFrame(c)            # julia dataframe again
```
This results in:
```
julia> c
1×1 PyPandasDataFrame
                        x
0 2023-04-13 14:36:13.939

julia> d
1×1 DataFrame
 Row │ x
     │ PyArray…
─────┼───────────────────────────────────
   1 │ UInt8[0xc0, 0x62, 0x31, 0x8c, 0x…
```
The same thing happens when initially defining `b` as a pandas dataframe, so the microsecond issue in #265 seems to not be the problem?   
```
julia> b = pd.DataFrame([[dt.datetime.now()]])
Python DataFrame:
                           0
0 2023-04-13 14:46:57.940077

julia> c = PyTable(b)
1×1 PyPandasDataFrame       
                           0
0 2023-04-13 14:46:57.940077

julia> d = DataFrame(c)
1×1 DataFrame
 Row │ 0
     │ PyArray…
─────┼───────────────────────────────────
   1 │ UInt8[0xc8, 0xf9, 0xa5, 0x7d, 0x…
```

cjdoris:
Just looking at this now. It all comes down to the fact that numpy arrays containing `numpy.datetime64` cannot be wrapped as a Julia array yet. Therefore it's falling back on converting each element, and because `numpy.datetime64` actually implements the buffer protocol with itemtype=1byte, that gets converted to a `PyArray{UInt8}` as you're seeing.

So there's two things that could be fixed here:
- Support wrapping arrays containing `datetime64`
- Support converting `datetime64` to something more useful than `PyArray{UInt8}`.

cjdoris:
(PS thanks for the report and the MWE)

hhaensel:
Just stumbled across that problem and solved it this way:

```julia
[...]

PR #334, opened 2023-07-01
Add support for datetime64

hhaensel:
This PR tentatively addresses #293

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -26,8 +26,9 @@ julia = "1.6.1"
 
 [extras]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
+DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 TestItemRunner = "f8b46487-2199-4994-9208-9a1283c18c0a"
 
 [targets]
-test = ["Aqua", "Test", "TestItemRunner"]
+test = ["Aqua", "Test", "TestItemRunner", "DataFrames"]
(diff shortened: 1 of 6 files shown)


hhaensel:
I think, this is still relevant.
@cjdoris if you think some modifications are necessary, please let me know.

cjdoris:
Thanks for the PR! A couple of comments:

PythonCall avoids relying on anything other than the Python standard library - so `Py(::Period)` should not be converted to a `numpy.timedelta64` (if anything it should be a `datetime.timdelta`). No objection to the reverse conversion though.

However, there are differing semantics going on. Both `datetime.timedelta` and `numpy.timedelta64` represent fixed periods of time, so 1 hour is always 3600 seconds. However a Julia `Dates.Period` represents a calendar period of time, whose precise length is context dependent, so for example 1 hour can be 3600 or 3601 seconds depending on if the hour has a leap second or not. Therefore it doesn't really make sense to convert between them. It might be OK to convert small units (seconds, milliseconds, etc) because these have a fixed length but (a) this feels like a special case and (b) maybe one day we'll add leap milliseconds to the calendar.

hhaensel:
> PythonCall avoids relying on anything other than the Python standard library - so Py(::Period) should not be converted to a numpy.timedelta64 (if anything it should be a datetime.timdelta). No objection to the reverse conversion though.

Sounds like a good proposal.

> However a Julia Dates.Period represents a calendar period of time

I see, I was not aware of the Period semantics in Julia, but it totally makes sense. I think as soon as we leave out `Year` everything should be fine. At least Julia converts happily between the lower units `Week`, `Hour`, `Day`, `Second`, etc ... while it throws an error as soon as `Year` is involved.

Should I wait for the refactoring, before continuing with this PR?

hhaensel:
Close in favour of #509

PR #365, opened 2023-09-20
Fix pyconvert cache update bugs by clearing the whole cache after any rule is added

### LINKED ISSUE #336: pyconvert_add_rule() ignored if cache has been populated
Currently, if I define a new rule from the beginning, everything works fine:
```julia
using PythonCall
using DataFrames

pdf = pytable(DataFrame(:a => 1:2, :b => 2:3))

PythonCall.pyconvert_rule_pandasdataframe(::Type{DataFrame}, x::Py) = DataFrame(PyTable(x))
PythonCall.pyconvert_add_rule("pandas.core.frame:DataFrame", DataFrame, PythonCall.pyconvert_rule_pandasdataframe, PythonCall.PYCONVERT_PRIORITY_ARRAY)

pyconvert(Any, pdf)
```
results in
```julia
2×2 DataFrame       
 Row │ a      b     
     │ Int64  Int64 
─────┼──────────────
   1 │     1      2
   2 │     2      3
```

However, if I do conversion once before the rule is defined, the conversion rules are cached and no conversion is performed. So far that's expected. But even if I delete the rule cache, the new rule is not applied.
```julia
using PythonCall
using DataFrames

pdf = pytable(DataFrame(:a => 1:2, :b => 2:3))

pyconvert(Any, pdf)

PythonCall.pyconvert_rule_pandasdataframe(::Type{DataFrame}, x::Py) = DataFrame(PyTable(x))
PythonCall.pyconvert_add_rule("pandas.core.frame:DataFrame", DataFrame, PythonCall.pyconvert_rule_pandasdataframe, PythonCall.PYCONVERT_PRIORITY_ARRAY)

empty!(PythonCall.PYCONVERT_RULES_CACHE)

pyconvert(Any, pdf)
```
results in
```julia
2×2 PyPandasDataFrame
   a  b
0  1  2
1  2  3
```
The reason is that somehow `@generated pyconvert_rules_cache` does not recalculate the value after the cache has been emptied.
```julia
julia> PythonCall.pyconvert_rules_cache(Any)
Dict{Ptr{PythonCall.C.PyObject}, Vector{Function}} with 1 entry:
  Ptr{PyObject} @0x00000219a6cf69e0 => [#54, #54, #54, #54, #54, #54, #54, #54, #54, #54, #54]
```
If I omit the `@generated` in front of `pyconvert_rules_cache`,  the new rules are considered.

If this is by design in order to gain performance, it's perhaps worth mentioning in the docs of `pyconvert_add_rule()`.  Otherwise removing `@generated` in [convert.jl](https://github.com/cjdoris/PythonCall.jl/blob/main/src/convert.jl#L284) could be a solution.

### LINKED ISSUE #364: pyconvert rule caching is buggy
**Affects:** PythonCall

**Describe the bug**
When I add a new conversion rule, it only applies when converting to Julia types `T` that have never had `pyconvert(T, x)` called on them.

```julia
julia> pyexec("""
       class Hello:
           pass
       """, Main)

julia> x = pyeval("Hello()", Main)
Python: <Hello object at 0x14759a210>

julia> pyconvert(Any, x)
Python: <Hello object at 0x14759a210>

julia> t = pytype(x)
Python: <class 'Hello'>

julia> PythonCall.pyconvert_add_rule(pyconvert(Any, t.__module__)*":"*pyconvert(Any, t.__qualname__), String, (_, _) -> "Hello!!") 

julia> pyconvert(Any, x)
Python: <Hello object at 0x14759a210> ### BUG ### <===========================================

julia> pyconvert(String, x)
"Hello!!"

julia> empty!.(values(PythonCall.PYCONVERT_RULES_CACHE)) ### A possible fix ###
2-element Vector{Dict{Ptr{PythonCall.C.PyObject}, Vector{Function}}}:
 Dict()
 Dict()

julia> pyconvert(Any, x)
"Hello!!"
```

**Your system**
PythonCall v0.9.14

**Possible fix**
 Make `pyconvert_add_rule` clear the rule cache

LilithHafner:
Fixes #364. Fixes #336.
It would be possible to do a more sophisticated partial cache clear, but IMO that would be premature optimization at this point.


test/convert.jl
```diff
@@ -222,3 +222,18 @@ end
     x1 = pyconvert(DateTime, pydatetime(2001, 2, 3, 4, 5, 6, 7000))
     @test x1 === DateTime(2001, 2, 3, 4, 5, 6, 7)
 end
+
+@testitem "pyconvert_add_rule (#364)" begin
```
cjdoris, line 226: Because of the stateful nature of the conversion rules, this test fails if you run it twice in succession in the same Julia session (e.g. from the VS Code testing pane). A fix is to use a random class name, such as `class = "Hello_364_$(string(rand(UInt), base=16))"`.

test/convert.jl
```diff
@@ -222,3 +222,18 @@ end
...
+    pyexec("""
+    class Hello_364:
+        pass
+    """, @__MODULE__)
+    x = pyeval("Hello_364()", @__MODULE__)
+    @test pyconvert(Any, x) === x
+    t = pytype(x)
+    PythonCall.pyconvert_add_rule(pyconvert(Any, t.__module__)*":"*pyconvert(Any, t.__qualname__), String, (_, _) -> "Hello!!")
```
cjdoris, line 234: Can simplify the class name to `"$(t.__module__):$(t.__qualname__)"`.

cjdoris (changes requested):
Great thanks! Just a couple of comments on the tests.

Fixes #336.

test/convert.jl
LilithHafner: Tehe, yes but only because of the test on line 232 :)

cjdoris:
Thank you!

PR #367, opened 2023-09-21
Switch argument order of pyconvert_add_rule

### LINKED ISSUE #363: Why does the argument order of `pyconvert_add_rule` deviate from convention
According to the [Julia Docs](https://docs.julialang.org/en/v1/manual/style-guide/#Write-functions-with-argument-ordering-similar-to-Julia-Base), function arguments should be placed first because "Putting a function argument first permits the use of [do](https://docs.julialang.org/en/v1/base/base/#do) blocks for passing multiline anonymous functions."

Is there a reason for not following this convention?
Would it be reasonable to support the conventional argument order in addition to the current argument order?
Can we switch fully to the conventional argument order and deprecate the current argument order / remove it in the next breaking release?

LilithHafner:
Fixes #363

### DIFF
--- a/src/convert.jl
+++ b/src/convert.jl
@@ -16,17 +16,17 @@ const PYCONVERT_RULES = Dict{String, Vector{PyConvertRule}}()
 const PYCONVERT_EXTRATYPES = Py[]
 
 """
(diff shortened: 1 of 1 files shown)


cjdoris:
Looks good thanks.

Will merge later - PythonCall is undergoing a refactor right now (for my future sanity) so I'm not merging PRs until that's done.

PR #380, opened 2023-10-10
Strip whitespace before parsing in seval

### LINKED ISSUE #379: `Main.seval` fails to strip whitespace and therefore throws when it shouldn't.
**Affects:** JuliaCall 

```py
>>> from juliacall import Main
>>> Main.seval("""
... function f()
...     0
... end
... """)
Julia: f (generic function with 1 method)
>>> if True:
...     Main.seval("""
...     function g()
...         1
...     end
...     """)
... 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "/Users/x/.julia/packages/PythonCall/qTEA1/src/jlwrap/module.jl", line 25, in seval
    return self._jl_callmethod($(pyjl_methodnum(pyjlmodule_seval)), expr)
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
juliacall.JuliaError: Base.Meta.ParseError("extra token after end of expression")
Stacktrace:
 [1] parse(str::String; raise::Bool, depwarn::Bool)
   @ Base.Meta ./meta.jl:272
 [2] parse
   @ ./meta.jl:266 [inlined]
 [3] pyjlmodule_seval(self::Module, expr::Py)
   @ PythonCall ~/.julia/packages/PythonCall/qTEA1/src/jlwrap/module.jl:13
 [4] _pyjl_callmethod(f::Any, self_::Ptr{PythonCall.C.PyObject}, args_::Ptr{PythonCall.C.PyObject}, nargs::Int64)
   @ PythonCall ~/.julia/packages/PythonCall/qTEA1/src/jlwrap/base.jl:62
 [5] _pyjl_callmethod(o::Ptr{PythonCall.C.PyObject}, args::Ptr{PythonCall.C.PyObject})
   @ PythonCall.C ~/.julia/packages/PythonCall/qTEA1/src/cpython/jlwrap.jl:47
```

The issue is that Python's multiline string parsing is different from Julia's and does not strip trailing spaces as much, so the strict behavior of Julia's `Meta.parse` does not make sense for strings that were written using Python's multiline string syntax.

I think PyCall doesn't have this issue, but I'm not positive.

I think that a good solution would be to strip inputs before parsing. It's hard to imagine a case when someone would want code with a leading or trailing space to throw a parse error.

LilithHafner:
Fixes #379 

I'm not sure how to add tests for this. I don't see tests from the Python side.

### DIFF
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -11,3 +11,9 @@ end
     @test s[1:2] == "ab"
     @test s[1:2:end] == "aaaab"
 end
+
+@testitem "seval" begin
+    m = Py(Main)
+    @test m.seval("1 + 1") === 2 # Basic behavior
+    @test m.seval("1 + 1\n ") === 2 # Trailing whitespace
+end
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -11,9 +11,3 @@ end
     @test s[1:2] == "ab"
     @test s[1:2:end] == "aaaab"
 end
-
-@testitem "seval" begin
-    m = Py(Main)
-    @test m.seval("1 + 1") === 2 # Basic behavior
-    @test m.seval("1 + 1\n ") === 2 # Trailing whitespace
-end
--- a/pytest/test_all.py
+++ b/pytest/test_all.py
@@ -1,2 +1,11 @@
 def test_import():
     import juliacall
+
+def test_seval():
+    from juliacall import Main
+
+    Main.seval("""
+    function f(x)
+        return x^2
+    end
+    """)
--- a/pytest/test_all.py
+++ b/pytest/test_all.py
@@ -1,11 +1,2 @@
 def test_import():
     import juliacall
-
-def test_seval():
-    from juliacall import Main
-
-    Main.seval("""
-    function f(x)
-        return x^2
-    end
-    """)
--- a/src/jlwrap/module.jl
+++ b/src/jlwrap/module.jl
@@ -10,7 +10,7 @@ function pyjlmodule_dir(self::Module)
 end
 
 function pyjlmodule_seval(self::Module, expr::Py)
-    Py(Base.eval(self, Meta.parse(pyconvert(String, expr))))
+    Py(Base.eval(self, Meta.parse(strip(pyconvert(String, expr)))))
 end
 
 function init_jlwrap_module()
--- a/test/jlwrap.jl
+++ b/test/jlwrap.jl
@@ -149,11 +149,11 @@ end
 end
 
 @testitem "base" begin
-    
+
 end
 
 @testitem "callback" begin
-    
+
 end
 
 @testitem "dict" begin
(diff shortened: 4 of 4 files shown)


cjdoris:
Thanks! You could use https://github.com/JuliaPy/PythonCall.jl/issues/379#issuecomment-1754652793 as the basis of some tests.

LilithHafner:
Okay, I put tests in `test/utils.jl` and tested this from the Julia side. Are there no CI tests with Python as the host language?

test/utils.jl
```diff
@@ -11,3 +11,9 @@ end
     @test s[1:2] == "ab"
     @test s[1:2:end] == "aaaab"
 end
+
+@testitem "seval" begin
+    m = Py(Main)
+    @test m.seval("1 + 1") === 2 # Basic behavior
+    @test m.seval("1 + 1\n ") === 2 # Trailing whitespace
+end
```
cjdoris, line 19: Could you move the test to https://github.com/JuliaPy/PythonCall.jl/blob/5f56a9b96b867a9f6742ab1d1e2361abd844e19f/test/jlwrap.jl#L189 please?

cjdoris (changes requested):

test/utils.jl
cjdoris: Also these tests are failing because `m.seval(...)` returns a `Py`.

cjdoris:
> Are there no CI tests with Python as the host language?

There are, under `pytest`. These get run in CI along with the Julia tests.

Currently all the Python-side tests do is check that `import juliacall` doesn't raise an exception.

LilithHafner:
My inclination would be to test this on the python side the way it came up in the wild, like this:

LilithHafner:
Where is `test_import` called from?

LilithHafner:
FWIW this is not a problem in Julia 1.10+, but it's still probably worth fixing here for compatibility with 1.6/1.9.

cjdoris:
> My inclination would be to test this on the python side the way it came up in the wild, like [this](https://github.com/JuliaPy/PythonCall.jl/pull/380/commits/bef8c6a5bd85daa771b01759409bc4cfccffcf4b)

I'd prefer the tests back in Julia, since that's where it's defined (it's more of a unit test that way).

cjdoris:
> Where is `test_import` called from?

It's called by magic. When you run the [`pytest`](https://docs.pytest.org/en/7.4.x/) program, it searches for functions called `test_something` and executes them.

LilithHafner:
So I'm assuming the new function I added is also called by magic. Is that test acceptable to you or would you like me to move it back to the Julia side?

cjdoris:
> So I'm assuming the new function I added is also called by magic. Is that test acceptable to you or would you like me to move it back to the Julia side?

Yes indeed. Back to Julia please - I'd like to keep the Python tests restricted to just the things extra that JuliaCall itself gives you.

cjdoris:
Thanks muchly!

Issue #384, opened 2023-10-19
`import juliacall` sometimes segfaults


LilithHafner:
**Affects:** JuliaCall

**Describe the bug**
I sometimes get segfaults when importing julicall in combination with other packages.

```
x@x ~ % python -c 'import juliacall; import torch; from torch.autograd import Function'
x@x ~ % python -c 'import torch; from torch.autograd import Function; import juliacall'
x@x ~ % python -c 'import torch; from torch.autograd import Function; import juliacall'
zsh: segmentation fault  python3 -c 
x@x ~ % python -c 'import torch; from torch.autograd import Function; import juliacall'
zsh: segmentation fault  python3 -c 
x@x ~ % python -c 'import torch; from torch.autograd import Function; import juliacall'
zsh: segmentation fault  python3 -c 
x@x ~ % python -c 'import torch; from torch.autograd import Function; import juliacall'
zsh: segmentation fault  python3 -c 
```

**Your system**
```
x@x ~ % python --version
Python 3.11.2
x@x ~ % julia -e 'using InteractiveUtils; versioninfo()'
Julia Version 1.9.3
Commit bed2cd540a1 (2023-08-24 14:43 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin22.4.0)
  CPU: 8 × Apple M2
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, apple-m1)
  Threads: 1 on 4 virtual cores
Environment:
  JULIA_EDITOR = code
  JULIA_PYTHONCALL_EXE = /Library/Frameworks/Python.framework/Versions/3.11/bin/python3
x@x ~ % python -m pip list                              
Package                   Version
------------------------- ---------
absl-py                   2.0.0
accelerate                0.23.0
aiohttp                   3.8.4
aiosignal                 1.3.1
astunparse                1.6.3
async-timeout             4.0.2
attrs                     22.2.0
cachetools                5.3.1
certifi                   2022.12.7
cffi                      1.15.1
charset-normalizer        3.1.0
click                     8.1.3
codespell                 2.2.6
comtypes                  1.1.14
contourpy                 1.0.7
cycler                    0.11.0
diffeqpy                  2.1.0
diffusers                 0.21.1
Faker                     19.10.0
filelock                  3.12.4
fire                      0.5.0
flatbuffers               23.5.26
[...]

PR #385, opened 2023-10-20
Add warning and hint when loading torch before juliacall

LilithHafner:
Before
```sh
x@x pythoncall % python -c 'import torch; from torch.autograd import Function; import juliacall'
zsh: segmentation fault  python3 -c 
x@x pythoncall % be sad
zsh: command not found: be
```
After
```sh
x@x pythoncall % python -c 'import torch; from torch.autograd import Function; import juliacall'
/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/juliacall/__init__.py:62: UserWarning: `torch` was loaded before the juliacall. This may cause a segfault. To avoid this, import `juliacall` *before* importing `torch`. For updates, see https://github.com/pytorch/pytorch/issues/78829
  warnings.warn(
zsh: segmentation fault  python3 -c 
x@x pythoncall % python -c 'import juliacall; import torch; from torch.autograd import Function'
x@x
```

Helps with https://github.com/JuliaPy/PythonCall.jl/issues/384, but doesn't solve the issue satisfactorily.

### DIFF
--- a/pysrc/juliacall/__init__.py
+++ b/pysrc/juliacall/__init__.py
@@ -53,6 +53,15 @@ def init():
     import ctypes as c
     import sys
     import subprocess
+    import warnings
+
+    # importing pytorch before juliacall sometimes causes segfaults. TODO: remove
+    if "torch" in sys.modules:
+        warnings.warn(
+            "`torch` was loaded before the juliacall. This may cause a segfault. "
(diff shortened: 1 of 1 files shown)


LilithHafner:
"Swap the import order and your segfaults will go away"

PR #444, opened 2024-01-26
Remove M1/M2 part from FAQ

MilesCranmer:
macOS M1/M2 is Tier 1: https://julialang.org/downloads/#supported_platforms so this PR removes the part from the FAQ about it being Tier 2.

(I also haven't encountered any issues with M1/M2 specifically, so I think it's on par in terms of reliability as the other platforms)


cjdoris:
Great thanks! Also it's good to know it works ok on these chips now - indeed I haven't had any bug reports specific to that for a long time.

PR #446, opened 2024-01-28
Update juliacall.md

MilesCranmer:


cjdoris:
Thanks for the correction!

cjdoris:
I agree it would be nice to do this automatically. The issue is that the installation of PythonCall is not handled by pip, it is handled by JuliaPkg (which JuliaCall depends on for package management). So JuliaPkg needs to somehow know that we are running the dev version of JuliaCall, and then switch behaviour accordingly. I don't know a robust way to check if a package is in "dev mode". I'm not even sure what that means.

An alternative would be to have juliapkg.json in the repo actually be the dev one, and change it whenever I make a release. But that would be extremely error prone.

A simple solution would simply to have an env var like `PYTHON_JULIAPKG_DEV=juliacall` (or Python flag `-X juliapkg-dev=juliacall`), and then for these packages JuliaPkg looks for `juliapkg-dev.json` instead of `juliapkg.json`.

PR #449, opened 2024-02-01
Fix around logo

hyrodium:
* Update logo on README
* Add favicon

**Before this PR**

image

image

**After this PR**

image

image


cjdoris:
Ah nice thanks! The high res logo shows up sharper on high density displays too.

PR #450, opened 2024-02-02
Add an example for `tname` for custom Python-to-Julia conversion rule

cuihantao:
I ran into the same issue as https://discourse.julialang.org/t/sympy-symbol-to-symbolics-jl-num-adding-a-rule/102915 and thus thought it may be helpful to add another example to the docstring.


cjdoris:
Thanks!

Issue #433, opened 2023-12-06
Main.seval can't parse multiple toplevel blocks


LilithHafner:
**Affects:** JuliaCall

```pycon
>>> from juliacall import Main
>>> Main.seval("x = 1+1; println(x)")
2
>>> Main.seval("x = 1+1\n println(x)")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/x/.julia/packages/PythonCall/qTEA1/src/jlwrap/module.jl", line 25, in seval
    return self._jl_callmethod($(pyjl_methodnum(pyjlmodule_seval)), expr)
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
juliacall.JuliaError: Base.Meta.ParseError("extra token after end of expression")
Stacktrace:
 [1] parse(str::String; raise::Bool, depwarn::Bool)
   @ Base.Meta ./meta.jl:272
 [2] parse
   @ ./meta.jl:266 [inlined]
 [3] pyjlmodule_seval(self::Module, expr::Py)
   @ PythonCall ~/.julia/packages/PythonCall/qTEA1/src/jlwrap/module.jl:13
 [4] _pyjl_callmethod(f::Any, self_::Ptr{PythonCall.C.PyObject}, args_::Ptr{PythonCall.C.PyObject}, nargs::Int64)
   @ PythonCall ~/.julia/packages/PythonCall/qTEA1/src/jlwrap/base.jl:62
 [5] _pyjl_callmethod(o::Ptr{PythonCall.C.PyObject}, args::Ptr{PythonCall.C.PyObject})
   @ PythonCall.C ~/.julia/packages/PythonCall/qTEA1/src/cpython/jlwrap.jl:47
```

The issue here, I suspect, is that JuiliaCall uses a parsing method that accepts only a single top-level block. This is not ideal as Julia's `eval ∘ Meta.parse` accepts any number top-level expressions. What makes this a bug, though, is how subtle the difference between a single top-level block and multiple top-level blocks is. Specifically `a; b` is okay but `a\nb` is not which is very bad. 

Fixing this may also fix #379 which would allow us to revert #380

cjdoris:
If we're being pedantic about it this isn't a bug: `a;b` really is a single block expression (this is the documented meaning of `;`) whereas `a\nb` is just two expressions separated by whitespace, which don't become a single expression until wrapped in some higher block such as `begin ... end`.

[...]

PR #452, opened 2024-02-08
Allow multi-block expressions in `seval`

MilesCranmer:
Simpler than I thought!

Fixes https://github.com/JuliaPy/PythonCall.jl/issues/433

@LilithHafner @cjdoris 

The only difference between these is that `Meta.parse` checks for multiple blocks and throws an error:

```julia
    ex, pos = parse(str, 1; filename, greedy=true, raise, depwarn)
    if isexpr(ex, :error)
        return ex
    end
    if pos <= ncodeunits(str)
        raise && throw(ParseError("extra token after end of expression"))
        return Expr(:error, "extra token after end of expression")
    end
    return ex
```

whereas `Meta.parseall` ignores the number of blocks:

```julia
    ex,_ = _parse_string(text, String(filename), lineno, 1, :all)
    return ex
```

Note that otherwise the behavior is identical.


LilithHafner:
It's worth noting that `Meta.parseall` is undocumented, but this still looks like a good fix to me.

I suppose it would be possible to do this with public API by repeatedly parsing with the two-arg parse method and then manually joining them into a single expr, but I like this approach better.

LilithHafner (approved):

MilesCranmer:
Is it public since it is `parseall` rather than `_parseall`? (I don't know; just wondering). There are even some exported functions from Julia that don't have docs IIRC.

I just tried to do a 2-arg Meta.parse version and ran into some sharp edges of the interface. For this reason I think `parseall` is safer even if not documented given it's unit-tested and calls the same underlying `_parse_str` but with a different option. So, I'm happy to hear you are on board too.

cjdoris:
Great thanks! Given `Meta.parseall` is used directly for code loading it seems unlikely to go away. Would be good to get it documented in Julia though, since there's no other documented reliable way to parse top-level expressions.

PR #456, opened 2024-02-10
Update faq.md with example on running JuliaCall inside Blender

cdsousa:


MilesCranmer:
Would it be okay to copy the example directly in the docs somewhere? Would help with discoverability.

cjdoris:
Thanks!

PR #461, opened 2024-02-12
Add PySR integration test

MilesCranmer:
Now that PySR has switched to PythonCall.jl, here is the promised integration test :)

I'm installing and running it in a separate virtualenv so as to not interfere with anything.

### DIFF
--- a/pytest/test_all.py
+++ b/pytest/test_all.py
@@ -17,3 +17,39 @@ def test_issue_394():
     assert jl.f is f
     assert jl.y is y
     assert jl.seval("f(x)") == 4
+
+
+def test_integration_pysr():
+    "Integration tests for PySR"
+    import subprocess
+    import sys
+    import tempfile
+
+    with tempfile.TemporaryDirectory() as tempdir:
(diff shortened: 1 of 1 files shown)


cjdoris:
This looks cool thanks! However I don't think integration tests are suitable for running on every commit in regular CI. Instead I think it would make sense to have a separate github workflow for integration tests which is run manually - and I'll ensure to run it before each release.

Are you familiar enough with github workflows to do this or would you rather I did it?

MilesCranmer:
Sounds good to me – just adding the `workflow_dispatch` key right? 

Also would it be useful to have a daily/weekly cron job or something, to test against `main`? (but not so regular as to be a nuisance)

cjdoris:
Yeah a weekly cron job would also be fine.

MilesCranmer:
@cjdoris I have greatly simplified the integration test so I think it is now appropriate to run on each commit. In fact it takes less time than many of the other GitHub actions.

I think it would be very useful to have something like this as it tests more complex workflows for scenarios not probed by the unittests alone, such as giving an early indication of issues such as #537 (due to the PyNULL typo).

What do you think?

PR #478, opened 2024-03-17
Unittest IPython extension + Automatic loading

MilesCranmer:
This adds unittesting for the IPython extension, using nbval: https://nbval.readthedocs.io/

As discussed in issue () with @cjdoris, I have also set up the IPython extension to automatically load when it detects the user has loaded the package from inside Jupyter. I've done this with PySR and it has been really nice as you can basically just do `import pysr` and then start typing out `%%julia` blocks of code as if you were using a native Julia notebook.

This can be switched off with a new `"autoload_extensions"` configuration parameter. The default behavior will be to assume `"yes"` but will also print a helpful message to the user letting them know how to disable it if they desire. Setting this parameter explicitly to `"yes"` will disable this message.


pytest/test_all.py
```diff
@@ -75,3 +75,21 @@ def test_issue_433():
...
+    import sys
+    from pathlib import Path
+
+    result = subprocess.run(
+        [
+            sys.executable,
+            "-m",
+            "pytest",
+            "--nbval",
+            str(Path(__file__).parent / "test_nb.ipynb"),
```
MilesCranmer, line 91: Not sure if coverage is propagated through recursive calls to `pytest`. If you would prefer we can just add a second `pytest` call to the CI but I slightly prefer calling it from within the existing tests so that a single `pytest` call tests everything.

cjdoris: You could instead simply add the `--nbval` arg to the `pytest` command in `tests.yml` - then it will find tests in both notebooks and *.py files.

pysrc/juliacall/__init__.py
```diff
@@ -248,8 +248,37 @@ def jlstr(x):
                 "PYTHON_JULIACALL_HANDLE_SIGNALS=no."
             )
 
-init()
+    # Next, automatically load the juliacall extension if we are in a Jupyter notebook
+    CONFIG['autoload_extensions'] = choice('autoload_extensions', ['yes', 'no'])[0]
```
cjdoris, line 252: Can we rename the option to `autoload_ipython_extension`? I know it's a mouthful but the current name doesn't really say what sort of extension is being loaded.

pysrc/juliacall/__init__.py
```diff
@@ -248,8 +248,37 @@ def jlstr(x):
...
+    if CONFIG['autoload_extensions'] in {'yes', None}:
+        try:
+            get_ipython = sys.modules['IPython'].get_ipython
+
+            if 'IPKernelApp' not in get_ipython().config:
```
cjdoris, line 257: What does this check do? Please add a comment.

pysrc/juliacall/__init__.py
```diff
@@ -248,8 +248,37 @@ def jlstr(x):
...
+                raise ImportError('console')
+
+            if CONFIG['autoload_extensions'] is None:
+                # Only let the user know if it was not explicitly set
+                print(
+                    "Detected Jupyter notebook. Loading juliacall extension. To disable, you can either "
```
cjdoris, line 263: The IPython kernel can be used in places other than Jupyter, so can you make the first bit more general please? Perhaps "IPython detected, loading juliacall extension."

cjdoris:
Thanks this looks good - could you resolve the merge conflicts please so CI can run?

MilesCranmer:
> Thanks this looks good - could you resolve the merge conflicts please so CI can run?

I was very confused about that error... It's actually up-to-date with `main`. Maybe some GitHub bug? Let's see if it goes away with new commits.

pysrc/juliacall/__init__.py
MilesCranmer: Done.

Btw, I learned about it from tqdm's automatic functionality: https://github.com/tqdm/tqdm/blob/master/tqdm/autonotebook.py. If Python ever updates how this works, tqdm is probably a good place to figure out what to do for this type of behavior.

pysrc/juliacall/__init__.py
MilesCranmer: The current autoloading only applies to Jupyter notebooks. Want me to make it work for IPython as well?

pysrc/juliacall/__init__.py
MilesCranmer: Sure, sounds good.

pysrc/juliacall/__init__.py
MilesCranmer: Edit: just added it for IPython as well. My only fear in making it more general is that there might be other types of IPython notebooks that break... But I guess we will see.

MilesCranmer:
Okay, all suggestions implemented. I made it work for IPython as well.

MilesCranmer:
My one other question is I fear the current warning is too verbose and a bit annoying:

```python
In [1]: from juliacall import Main as jl
Detected IPython. Loading juliacall extension. To disable, you c
an either set the environment variable PYTHON_JULIACALL_AUTOLOAD
_IPYTHON_EXTENSION=no or pass the command line argument '-X juli
acall-autoload-ipython-extension=no'. Inside Jupyter, you can do
 this with `import os; os.environ['PYTHON_JULIACALL_AUTOLOAD_IPY
THON_EXTENSION'] = 'no'`. To suppress this message, set PYTHON_J
ULIACALL_AUTOLOAD_IPYTHON_EXTENSION=yes.
```

Imagine getting hit with this every time you load the package...

I propose we instead just make this:

```python
In [1]: from juliacall import Main as jl
Detected IPython. Loading juliacall extension.
```

and include the other information in the docs. Users can google how to turn it off if they want, and will land on the docs page with more details.

pytest/test_all.py
MilesCranmer: Oh, I didn't even realize it would combine them. Cool!

MilesCranmer:
All suggestions implemented and I cleaned up the commit history

pysrc/juliacall/__init__.py
```diff
@@ -248,8 +248,34 @@ def jlstr(x):
                 "PYTHON_JULIACALL_HANDLE_SIGNALS=no."
             )
 
-init()
+    # Next, automatically load the juliacall extension if we are in IPython or Jupyter
+    CONFIG['autoload_ipython_extension'] = choice('autoload_ipython_extension', ['yes', 'no'])[0]
```
cjdoris, line 252: Could you put this near the top of init? Just so if it's set incorrectly the error gets raised early.

pysrc/juliacall/__init__.py
```diff
@@ -248,8 +248,34 @@ def jlstr(x):
...
+    if CONFIG['autoload_ipython_extension'] in {'yes', None}:
+        try:
+            get_ipython = sys.modules['IPython'].get_ipython
+
+            if CONFIG['autoload_ipython_extension'] is None:
+                # Only let the user know if it was not explicitly set
+                print(
```
cjdoris, line 259: Should maybe be a warning, so can be disabled using Python's warnings mechanisms?

pysrc/juliacall/__init__.py
```diff
@@ -248,8 +248,34 @@ def jlstr(x):
...
+                    "Detected IPython. Loading juliacall extension. To disable, you can either "
```
cjdoris, line 260: I see what you mean about this being quite long.

Most of this should be in the docs so please add it there - there's already a config bit in the JuliaCall page.

Then you could replace most of this message with a hyperlink to the docs?

cjdoris (changes requested):
The mechanism looks great, just a couple of small comments to address then it's good to go.

pysrc/juliacall/__init__.py
cjdoris: I'm ambivalent about this.

pysrc/juliacall/__init__.py
MilesCranmer: Don't we want it come after Julia is initialized? Otherwise we would see the printout for installing Julia, and the extension autoload message would be lost above it.

pysrc/juliacall/__init__.py
MilesCranmer: If it is a warning, it feels like you *need* to set the environment variable (one way or the other), and ends up being a pain. When I have this set up I just want it to run automatically and not worry about it. A single-line print seems nice though.

I also tried `logger.info` but it's also a pain because you need to reconfigure logging defaults... I guess I think this is really what `print` is made for. The user can always set that env variable to turn off printing.

pysrc/juliacall/__init__.py
MilesCranmer: Done!

MilesCranmer:
Okay I took the advice into account and added docs on both the environment variables page and the IPython page. The automatic loading now links to the IPython page (https://juliapy.github.io/PythonCall.jl/stable/compat/#IPython) which seems nice as it points the user to the API. It's great for people who learn by trying.

cjdoris:
Thanks muchly :)

PR #494, opened 2024-04-24
making it possible to use another package instead of `pickle` for serialization (e.g. `dill`)

schlichtanders:
Looking into the documentation it looks like PythonCall uses Environment Variables for configuration.

Hence I added a simple `"JULIA_PYTHONCALL_PICKLE"` Environment Variable which defaults to `"pickle"`. This way users like me can easily use a more complete serialization system, as "pickle" cannot serialize everything (e.g. precompiled code, which is my usecase).


cjdoris:
Thanks!

PR #509, opened 2024-06-16
Add timedelta, timedelta64 and datetime64 plus respective conversions

### LINKED ISSUE #293: `DataFrame(::PyPandasDataFrame)` converts date & datetime to bytes

hhaensel:
This PR replaces #334 and takes into account the major refactoring of PythonCall.
Particularly, it fixes #293.

#### Python Constructors
```julia
julia> pytimedelta(hour = 1, minute = 2)
Python: datetime.timedelta(seconds=3720)

julia> pytimedelta64(hour = 1, minute = 2)
Python: numpy.timedelta64(62,'m')

julia> pytimedelta64(year = 2, month = 3)
Python: numpy.timedelta64(27,'M')

julia> pydatetime64(year = 2024, month = 3)
Python: numpy.datetime64('2024-03-01T00:00:00')
```
#### DataFrame handling
```julia
julia> jdf = DataFrame(x = [now() + Second(rand(1:1000)) for _ in 1:100], y = [Second(n) for n in 1:100])
100×2 DataFrame
 Row │ x                        y
     │ DateTime                 Second
─────┼──────────────────────────────────────
   1 │ 2024-06-17T00:31:31.236  1 second
   2 │ 2024-06-17T00:30:30.236  2 seconds
   3 │ 2024-06-17T00:41:22.236  3 seconds
  ⋮  │            ⋮                  ⋮
  98 │ 2024-06-17T00:36:05.236  98 seconds
  99 │ 2024-06-17T00:38:38.236  99 seconds
 100 │ 2024-06-17T00:28:21.236  100 seconds
                             94 rows omitted

julia> pdf = pytable(jdf)
Python:
                         x               y
0  2024-06-17 00:31:31.236 0 days 00:00:01
1  2024-06-17 00:30:30.236 0 days 00:00:02
2  2024-06-17 00:41:22.236 0 days 00:00:03
3  2024-06-17 00:33:52.236 0 days 00:00:04
           ... 4 more lines ...
97 2024-06-17 00:36:05.236 0 days 00:01:38
98 2024-06-17 00:38:38.236 0 days 00:01:39
99 2024-06-17 00:28:21.236 0 days 00:01:40

[100 rows x 2 columns]

julia> DataFrame(PyTable(pdf))
100×2 DataFrame
 Row │ x                        y
     │ DateTime                 Compound…
─────┼───────────────────────────────────────────────
   1 │ 2024-06-17T00:31:31.236  1 second
   2 │ 2024-06-17T00:30:30.236  2 seconds
   3 │ 2024-06-17T00:41:22.236  3 seconds
  ⋮  │            ⋮                      ⋮
  98 │ 2024-06-17T00:36:05.236  1 minute, 38 seconds
  99 │ 2024-06-17T00:38:38.236  1 minute, 39 seconds
 100 │ 2024-06-17T00:28:21.236  1 minute, 40 seconds
                                      94 rows omitted
```
### Background
I chose to use `Dates.CompoundPeriod` as result type of default conversion from timedelta64 as both types support year, month and minor period units.
```julia
julia> pyconvert(Any, pytimedelta64(Year(2)))
2 years
```
Both Python and Julia do not convert between Year/Month and the other period types, so there is no danger with this choice to arrive at ill-determined intervals.
The difference is that Julia allows addition/subtraction of mixed types while Python/Numpy throws an error.

The difference to the previous PR is that all conversions rely on either builtin or numpy functions and do not use pandas.

During developing the conversion I found a very strange error on my Window machine; numpy's `datetime_data()` correctly reports the unit but not the value, so I had to convert the value myself.
```julia
julia> x = pydatetime64(1970, 1, 2, 3, 4, 5)
Python: numpy.datetime64('1970-01-02T03:04:05')

julia> pyimport("numpy").datetime_data(x)
Python: ('s', 1)
```

Ordering of arguments for `pytimedelta()` was chosen to be identical to the python version, while ordering for `pytimedelta64()` is strictly descending, except `week` which comes last.


hhaensel:
@cjdoris what do you think about this PR?

hhaensel:
@cjdoris still hoping to get this integrated ...
Is there anything pending or unclear where I could support?

MilesCranmer:
Just chiming in with some comments:

1: Could you add some tests for this? This adds a lot of new features so I think should have testing to cover everything. Ideally the test coverage of the diff should be 100%.

2: Is the 3-arg version of the `@py` needed? I don't see it being used anywhere. I think it should go into a different PR in any case.

src/Convert/numpy.jl
```diff
@@ -27,6 +27,66 @@ const NUMPY_SIMPLE_TYPES = [
     ("complex128", ComplexF64),
 ]
 
+function pydatetime64(
+    _year::Int=0, _month::Int=1, _day::Int=1, _hour::Int=0, _minute::Int=0,_second::Int=0, _millisecond::Int=0, _microsecond::Int=0, _nanosecond::Int=0;
+    year::Int=_year, month::Int=_month, day::Int=_day, hour::Int=_hour, minute::Int=_minute, second::Int=_second,
+    millisecond::Int=_millisecond, microsecond::Int=_microsecond, nanosecond::Int=_nanosecond
```
MilesCranmer, line 33: Just wondering, why not `::Integer`?

src/Convert/numpy.jl
```diff
@@ -27,6 +27,66 @@ const NUMPY_SIMPLE_TYPES = [
...
+)
+    pyimport("numpy").datetime64("$(DateTime(year, month, day, hour, minute, second))") + pytimedelta64(;millisecond, microsecond, nanosecond)
+end
+function pydatetime64(@nospecialize(x::T)) where T <: Period
+    T <: Union{Week, Day, Hour, Minute, Second, Millisecond, Microsecond} || 
+        error("Unsupported Period type: ", "Year, Month and Nanosecond are not supported, consider using pytimedelta64 instead.")
```
MilesCranmer, line 39:
```suggestion
        error("Unsupported Period type: `$x::$T`. Consider using pytimedelta64 instead.")
```

src/Convert/numpy.jl
```diff
@@ -27,6 +27,66 @@ const NUMPY_SIMPLE_TYPES = [
...
+    args = T .== (Day, Second, Millisecond, Microsecond,  Minute, Hour, Week)
```
MilesCranmer, line 40:
```suggestion
    args = map(Base.Fix1(isa, x), (Day, Second, Millisecond, Microsecond,  Minute, Hour, Week))
```

Since `==` is a different operator than `<:` as used in L38. If `Day` were to have a parametric type, this behavior would be different.

src/Convert/numpy.jl
```diff
@@ -27,6 +27,66 @@ const NUMPY_SIMPLE_TYPES = [
...
+    pydatetime64(x.value .* args...)
```
MilesCranmer, line 41:
```suggestion
    pydatetime64(map(Base.Fix1(*, x.value), args)...)
```

src/Convert/numpy.jl
```diff
@@ -27,6 +27,66 @@ const NUMPY_SIMPLE_TYPES = [
     ("complex128", ComplexF64),
 ]
 
+function pydatetime64(
+    _year::Int=0, _month::Int=1, _day::Int=1, _hour::Int=0, _minute::Int=0,_second::Int=0, _millisecond::Int=0, _microsecond::Int=0, _nanosecond::Int=0;
+    year::Int=_year, month::Int=_month, day::Int=_day, hour::Int=_hour, minute::Int=_minute, second::Int=_second,
+    millisecond::Int=_millisecond, microsecond::Int=_microsecond, nanosecond::Int=_nanosecond
+)
+    pyimport("numpy").datetime64("$(DateTime(year, month, day, hour, minute, second))") + pytimedelta64(;millisecond, microsecond, nanosecond)
```
MilesCranmer, line 35: Is `pyimport("numpy")` the correct API call, or is that just to be used in user packages?

src/Convert/numpy.jl
```diff
@@ -27,6 +27,66 @@ const NUMPY_SIMPLE_TYPES = [
...
+    isempty(x.periods) ? pytimedelta64(Second(0)) : sum(pytimedelta64.(x.periods))
+end
+export pytimedelta64
+
+function pyconvert_rule_datetime64(::Type{DateTime}, x::Py)
+    unit, count = pyconvert(Tuple, pyimport("numpy").datetime_data(x))
+    value = reinterpret(Int64, pyconvert(Vector, x))[1]
+    units = ("Y", "M", "W", "D", "h", "m", "s", "ms", "us", "ns")
+    types = (Year, Month, Week, Day, Hour, Minute, Second, Millisecond, Microsecond, Nanosecond)
+    T = types[findfirst(==(unit), units)]
```
MilesCranmer, line 73:
```suggestion
    T = types[findfirst(==(unit), units)::Int]
```

src/Convert/numpy.jl
```diff
@@ -27,6 +27,66 @@ const NUMPY_SIMPLE_TYPES = [
...
+    pyconvert_return(DateTime(_base_datetime) + T(value * count))
+end
+
+function pyconvert_rule_timedelta64(::Type{CompoundPeriod}, x::Py)
+    unit, count = pyconvert(Tuple, pyimport("numpy").datetime_data(x))
+    value = reinterpret(Int64, pyconvert(Vector, x))[1]
+    units = ("Y", "M", "W", "D", "h", "m", "s", "ms", "us", "ns")
+    types = (Year, Month, Week, Day, Hour, Minute, Second, Millisecond, Microsecond, Nanosecond)
+    T = types[findfirst(==(unit), units)]
```
MilesCranmer, line 82:
```suggestion
    T = types[findfirst(==(unit), units)::Int]
```

src/Convert/numpy.jl
```diff
@@ -27,6 +27,66 @@ const NUMPY_SIMPLE_TYPES = [
...
+    pyconvert_return(CompoundPeriod(T(value * count)) |> canonicalize)
```
MilesCranmer, line 83: The proper way to do this would be to use `Base.Cartesian.@nif`. That way you could write this code to avoid dynamic dispatch on `types` (which will be very slow).

src/Convert/numpy.jl
```diff
@@ -27,6 +27,66 @@ const NUMPY_SIMPLE_TYPES = [
...
+end
+export pytimedelta64
+
+function pyconvert_rule_datetime64(::Type{DateTime}, x::Py)
+    unit, count = pyconvert(Tuple, pyimport("numpy").datetime_data(x))
+    value = reinterpret(Int64, pyconvert(Vector, x))[1]
+    units = ("Y", "M", "W", "D", "h", "m", "s", "ms", "us", "ns")
+    types = (Year, Month, Week, Day, Hour, Minute, Second, Millisecond, Microsecond, Nanosecond)
+    T = types[findfirst(==(unit), units)]
+    pyconvert_return(DateTime(_base_datetime) + T(value * count))
```
MilesCranmer, line 74: Similar to other comment – you should write this using `Base.Cartesian.@nif` over the `types` tuple to avoid dynamic dispatch.

src/Convert/numpy.jl
```diff
@@ -27,6 +27,66 @@ const NUMPY_SIMPLE_TYPES = [
...
+function pytimedelta64(
+    _year::Int=0, _month::Int=0, _day::Int=0, _hour::Int=0, _minute::Int=0, _second::Int=0, _millisecond::Int=0, _microsecond::Int=0, _nanosecond::Int=0, _week::Int=0;
+    year::Int=_year, month::Int=_month, day::Int=_day, hour::Int=_hour, minute::Int=_minute, second::Int=_second, microsecond::Int=_microsecond, millisecond::Int=_millisecond, nanosecond::Int=_nanosecond, week::Int=_week)
+    pytimedelta64(sum((
+        Year(year), Month(month), # you cannot mix year or month with any of the below units in python, the error will be thrown by `pytimedelta64(::CompoundPeriod)`
+        Day(day), Hour(hour), Minute(minute), Second(second), Millisecond(millisecond), Microsecond(microsecond), Nanosecond(nanosecond), Week(week))
+    ))
+end
+function pytimedelta64(@nospecialize(x::T)) where T <: Period
+    index = findfirst(==(T), (Year, Month, Week, Day, Hour, Minute, Second, Millisecond, Microsecond, Nanosecond, T))
```
MilesCranmer, line 58:
```suggestion
    index = findfirst(==(T), (Year, Month, Week, Day, Hour, Minute, Second, Millisecond, Microsecond, Nanosecond, T))::Int
```
to avoid dynamic dispatch

src/Convert/numpy.jl
```diff
@@ -27,6 +27,66 @@ const NUMPY_SIMPLE_TYPES = [
...
+end
+function pydatetime64(@nospecialize(x::T)) where T <: Period
+    T <: Union{Week, Day, Hour, Minute, Second, Millisecond, Microsecond} || 
+        error("Unsupported Period type: ", "Year, Month and Nanosecond are not supported, consider using pytimedelta64 instead.")
+    args = T .== (Day, Second, Millisecond, Microsecond,  Minute, Hour, Week)
+    pydatetime64(x.value .* args...)
+end
+function pydatetime64(x::CompoundPeriod)
+    x =  canonicalize(x)
+    isempty(x.periods) ? pydatetime64(Second(0)) : sum(pydatetime64.(x.periods))
```
MilesCranmer, line 45:
```suggestion
    isempty(x.periods) ? pydatetime64(Second(0)) : sum(pydatetime64, x.periods)
```

src/Convert/numpy.jl
```diff
@@ -27,6 +27,66 @@ const NUMPY_SIMPLE_TYPES = [
...
+end
+export pydatetime64
+
+function pytimedelta64(
+    _year::Int=0, _month::Int=0, _day::Int=0, _hour::Int=0, _minute::Int=0, _second::Int=0, _millisecond::Int=0, _microsecond::Int=0, _nanosecond::Int=0, _week::Int=0;
+    year::Int=_year, month::Int=_month, day::Int=_day, hour::Int=_hour, minute::Int=_minute, second::Int=_second, microsecond::Int=_microsecond, millisecond::Int=_millisecond, nanosecond::Int=_nanosecond, week::Int=_week)
+    pytimedelta64(sum((
+        Year(year), Month(month), # you cannot mix year or month with any of the below units in python, the error will be thrown by `pytimedelta64(::CompoundPeriod)`
```
MilesCranmer, line 53: This comment
> you cannot mix year or month with any of the below units in python, the error will be thrown by `pytimedelta64(::CompoundPeriod)`
Should be presented to the user as a descriptive error message rather than a comment in the function

src/Convert/numpy.jl
```diff
@@ -27,6 +27,66 @@ const NUMPY_SIMPLE_TYPES = [
...
+    value = reinterpret(Int64, pyconvert(Vector, x))[1]
+    units = ("Y", "M", "W", "D", "h", "m", "s", "ms", "us", "ns")
+    types = (Year, Month, Week, Day, Hour, Minute, Second, Millisecond, Microsecond, Nanosecond)
+    T = types[findfirst(==(unit), units)]
+    pyconvert_return(DateTime(_base_datetime) + T(value * count))
+end
+
+function pyconvert_rule_timedelta64(::Type{CompoundPeriod}, x::Py)
+    unit, count = pyconvert(Tuple, pyimport("numpy").datetime_data(x))
+    value = reinterpret(Int64, pyconvert(Vector, x))[1]
```
MilesCranmer, line 79: Is `reinterpret` safe here? Is there a better alternative to use?

src/Convert/numpy.jl
```diff
@@ -54,4 +114,14 @@ function init_numpy()
         iscomplex && pyconvert_add_rule(name, Complex, rule)
         isnumber && pyconvert_add_rule(name, Number, rule)
     end
+
+    priority = PYCONVERT_PRIORITY_ARRAY
+    pyconvert_add_rule("numpy:datetime64", DateTime, pyconvert_rule_datetime64, priority)
+    for T in (CompoundPeriod, Year, Month, Day, Hour, Minute, Second, Millisecond, Microsecond, Nanosecond, Week)
+        pyconvert_add_rule("numpy:timedelta64", T, pyconvert_rule_timedelta64, priority)
+    end
```
MilesCranmer, line 122: Since Julia is unlikely to unroll this loop, you should use `Base.Cartesian.@nexpr` here to avoid dynamic dispatch.

src/Convert/numpy.jl
MilesCranmer, line 41: Probably better to rewrite this with `Base.Cartesian.@nif` rather than doing a masked sum, since you know there will be only 1 element in the sum.

src/Convert/numpy.jl
hhaensel: Integer sounds more reasonable, indeed.
I shall change that.

hhaensel:
Thanks for all the comments, I already learned some new things about Julia, which is always nice.

I will go through them the next days and once we've agreed on the solutions write the test functions.

Just want to re-raise my question from the previous PR.

Do you think, we should stick with CompoundPeriod as default conversion type?
Alternatively, we could go for Period instead.

src/Convert/numpy.jl
hhaensel: I saw similar calls at different places in the package, so I took this approach. But I also wouldn't know how to code a timedelta64 without calling pyimport.
Please let me know if there's a better solution.

src/Convert/numpy.jl
hhaensel: This sum is dammed fast (16ns), and I couldn't beat it with a different version.

src/Convert/numpy.jl
hhaensel: Maybe the comment isn't clear enough.
Python throws a well understandable descriptive error in case of wrong usage, so no need for us to do so. Agree?

src/Convert/numpy.jl
hhaensel: Again, tested with a julia function calls and found that the julia part is around 25ns, whereas the python call is around 1.5microsecond

src/Convert/numpy.jl
hhaensel: I thought, pyconvert creates a new Julia Vector which is not mapped onto Python data. If that would be the case, we'd need to wrap the vector by a `copy()`.

src/Convert/numpy.jl
hhaensel: Tried my best, but I'm not sure how to test whether this will speed up things

hhaensel:
Concerning the three-argument version of `@py`; it's probably best to remove it as it goes against the authors' initial idea of being as close as possible to Julia syntax, right?

hhaensel:
@cjdoris @MilesCranmer 
Apart from my question above this PR would be ready to merge from my end.
Please let me know if you have other concerns.
We may want to add testing, please also let me know.

hhaensel:
@cjdoris @MilesCranmer I'd really love to see these changes merged. Please let me know if there are any remaining things to be done from your end?

MilesCranmer:
Could you please add some tests and docs?

MilesCranmer:
Can you rebase on #583 to fix the tests?

hhaensel:
I did rebase, the errors come from the new julia version.
I already submitted one PR on pushfirst!, there's another issue with the doc function that I cannot reproduce on my laptop.

MilesCranmer:
I see, thanks. In the meantime can you add tests and docs for the code changes? Ideally we should have 100% test coverage.

hhaensel:
I did add tests for pytimedelta.
Tests of pytimedelta64 depend on numpy. But as far as I could see, you did not define a CondaPkg environment so that numpy or pandas could be tested. Moreover, I saw that none of the standard python API is documented by docstrings, e.g. `pytime()`.
Therefore, I am a bit lost where to begin adding docstrings for my extensions.
I tested using
```julia
using CondaPk
CondaPkg.add("pandas")
```
which I think is a reasonable approach.
If you are of the same opinion, I could write a PR that fixes the two locations, where :pandas is missing in the current tests and then continue completing this PR so that finally it doesn't error anymore.

MilesCranmer:
Were the test files not committed? I don’t see any added tests

hhaensel:
I submitted #588 and #589. I propose to merge these first before continuing here with docstrings and tests.

hhaensel:
When writing the tests for pytimedelta and pytimedelta64 I realised that my choice for keyword argument naming might be confusing for users.
Currently it is
```julia
pytimedelta(second = 1)
```
but
```julia
pyimport("datetime").timedelta(seconds = 1)
```
My choice was inspired from the Dates API (e.g. `Second(1)`)
I tend to append an 's' to the keywords, what do you think?

hhaensel:
Changed the keywords. Tests should pass as soon as the other too PRs are merged.
One missing error is Aqua's complaining about REPL being a stale dependency.

hhaensel:
Added functionality that the unit of pytimedelta64() is defined by the least unit specified, also for the case 0.
Moreover, the user can specify whether time difference should be canonicalized, default is false.
So
```julia-repl
julia> pytimedelta64(Year(0))
Python: np.timedelta64(0,'Y')

julia> pytimedelta64(Week(0))
Python: np.timedelta64(0,'W')

julia> pytimedelta64(Millisecond(0))
Python: np.timedelta64(0,'ms')
```
and
```julia-repl
julia> pyconvert(Dates.CompoundPeriod, pytimedelta64(Second(60)))
60 seconds

julia> pyconvert(Dates.CompoundPeriod, pytimedelta64(Second(60), canonicalize = true))
1 minute
```
and
```julia-repl
julia> pyconvert(Dates.CompoundPeriod, pytimedelta64(Second(60)))
60 seconds

julia> PythonCall.Convert.CANONICALIZE_TIMEDELTA64[] = true
true

julia> pyconvert(Dates.CompoundPeriod, pytimedelta64(Second(60)))
1 minute
```

hhaensel:
I hope you find these changes consistent.

hhaensel:
There's an error for CondaPkg on Windows, but in principle, the tests pass.
Will try to find a better method for CondaPkg environment.

hhaensel:
I cherry-picked #589 in order to proceed with testing.
Now everything seems fixed for julia 1.11
- added tests for pytimedelta, pytimedelta64 and pydatetime64
- several fixes that came up during testing
- CondaPkg environment seems reasonably solved by moving to a test/Project.toml with its own CondaPkg.toml

The docs are probably the only remaining topic.
Should I simply add docstrings? What is your recommendation?

hhaensel:
@cjdoris @MilesCranmer Ping!

Do you think we can finalize this PR?

Just pushed two (final?) commits so that we now have
- removed `@py` method for python style imports (`from datetime import datetime as dt`) as there is already the julia-style import
- reordered the arguments of `pytimedelta()` in decreasing order (except week, which omes last), while the original Python `datetime.timedelta()` function has a very unexpected signature that probably no one uses (`pytimedeltatimedelta(days, seconds, microseconds, milliseconds, minutes, hours, weeks)`). This way `pytimedelta()` has a similar signature to `pytimedelta64()`
- tests included
- docstrings included

MilesCranmer:
Would need @cjdoris to review first. I know he has been very busy lately, but I am sure he will review it as soon as he has time

PR #516, opened 2024-07-01
call jl_atexit_hook when atexit

### LINKED ISSUE #515: finalizer may never be called
**Is your feature request related to a problem? Please describe.**
In Julia, it looks like [finalizers](https://docs.julialang.org/en/v1/base/base/#Base.finalizer) are always called when exiting.
E.g., the following code always prints "hola".
```julia
mutable struct S
    a
end
function S()
    return finalizer(_->Core.println("hola"), S(3))
end
S()
```

However, if we wrap it with juliacall, the finalizer is never executed.
```python
import juliacall
juliacall.Main.include("kk.jl")

```

**Describe the solution you'd like**
I'd like finalizers to always be executed

**Describe alternatives you've considered**
My current workaround is finishing with 
```juliacall.Main.GC.gc()
juliacall.Main.GC.gc()

```

dpinol:
Fixes #515

### DIFF
--- a/pysrc/juliacall/__init__.py
+++ b/pysrc/juliacall/__init__.py
@@ -3,6 +3,8 @@
 
 __version__ = '0.9.20'
 
+import atexit
+
 _newmodule = None
 
 def newmodule(name):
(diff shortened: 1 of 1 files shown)


cjdoris:
Cool thanks!

PR #520, opened 2024-07-10
make python GC always run on thread 1

### LINKED ISSUE #201: Segfaults with frequent python calls from thread 1 on multithreaded Julia
In a complex application with high-frequency python calls I always run into

```
signal (11): Segmentation fault
in expression starting at REPL[41]:2
_PyTrash_begin at /opt/conda/conda-bld/python-split_1649141344976/work/Objects/object.c:2125
```

segfaults when Julia has multiple threads (e.g. `JULIA_NUM_THREADS=4`). I do make sure to only use PythonCall from thread 1 (and I use asserts to verify that) but somehow that doesn't seem to be enough. When running with `JULIA_NUM_THREADS=1`, all is fine. Is PythonCall incompatible with multi-threaded Julia in general?

ericphanson:
closes #201 

Here, we make the GC queue threadsafe (by using a Channel), so we can safely enqueue objects from any thread. Then we process it from a dedicated task which runs on thread 1, which is the thread in which we have initalized the correct state in the python runtime (https://docs.python.org/3/c-api/init.html#non-python-created-threads). We continue to respect `GC.enable()` and `GC.disable()` but it should not be very useful anymore since it is no longer required to disable GC in threaded regions. (However users must still only interact with the python runtime via thread 1, unless they setup the right state).

This was a bit complicated to test unfortunately, but if I have reasoned correctly (if!) there should be no races in the testing.

Notes:

- I'm not sure if we should skip the task creation during `__init__` during precompilation (will it hang precompilation? is it fine to not GC then)?
- I'm not sure if this will be less performant in the single-threaded case, if we are not eagerly freeing but instead deferring to a separate task to do so
- I haven't used `Threads.Condition` or `Threads.Event` before, hopefully I got it right

### DIFF
--- a/src/GC/GC.jl
+++ b/src/GC/GC.jl
@@ -104,7 +104,9 @@ function launch_gc_task()
     # ensure the task runs from thread 1
     ccall(:jl_set_task_tid, Cvoid, (Any, Cint), task, 0)
     schedule(task)
-    Base.errormonitor(task)
+    if isdefined(Base, :errormonitor)
+        Base.errormonitor(task)
+    end
     GC_TASK[] = task
     task
 end
--- a/src/GC/GC.jl
+++ b/src/GC/GC.jl
@@ -86,12 +86,7 @@ function gc_loop()
         if ENABLED[] && !isempty(QUEUE)
             unsafe_process_queue!()
             # just for testing purposes
-            lock(GC_FINISHED)
-            try
-                notify(GC_FINISHED) 
-            finally
-                unlock(GC_FINISHED)
-            end
+            Base.@lock GC_FINISHED notify(GC_FINISHED)
         end
         # wait until there is both something to process
         # and GC is `enabled`
--- a/src/GC/GC.jl
+++ b/src/GC/GC.jl
@@ -105,8 +105,10 @@ end
 function enqueue(ptr::C.PyPtr)
     # if we are on thread 1:
     f = () -> begin
-        if ptr != C.PyNULL
-            C.Py_DecRef(ptr)
+        C.with_gil(false) do
+            if ptr != C.PyNULL
+                C.Py_DecRef(ptr)
+            end
         end
     end
     # otherwise:
@@ -121,9 +123,11 @@ end
 function enqueue_all(ptrs)
     # if we are on thread 1:
     f = () -> begin
-        for ptr in ptrs
-            if ptr != C.PyNULL
-                C.Py_DecRef(ptr)
+        C.with_gil(false) do
+            for ptr in ptrs
+                if ptr != C.PyNULL
+                    C.Py_DecRef(ptr)
+                end
             end
         end
     end
--- a/src/GC/GC.jl
+++ b/src/GC/GC.jl
@@ -66,6 +66,8 @@ function enable()
     return
 end
 
+# This is called within a finalizer so we must not task switch
+# (so no printing nor blocking on Julia-side locks)
 function enqueue_wrapper(f, g)
     t = @elapsed begin
         if C.CTX.is_initialized
@@ -81,9 +83,6 @@ function enqueue_wrapper(f, g)
                 end
                 if Threads.threadid() == 1
                     f()
-                    # if ptr != C.PyNULL
-                    # C.Py_DecRef(ptr)
-                    # end
                     handled = true
                 end
                 if !old_sticky
@@ -92,9 +91,6 @@ function enqueue_wrapper(f, g)
             end
             if !handled
                 g()
-                # if ptr != C.PyNULL
-                # put!(QUEUE, ptr)
-                # end
             end
         end
     end
@@ -141,44 +137,6 @@ function enqueue_all(ptrs)
     end
     enqueue_wrapper(f, g)
 end
-
-# function enqueue_all(ptrs)
-#     t = @elapsed begin
-#         if C.CTX.is_initialized
-#             # Eager path: if we are already on thread 1,
-#             # we eagerly decrement
-#             handled = false
-#             if ENABLED[] && Threads.threadid() == 1
-#                 # temporarily disable thread migration to be sure
-#                 # we call `C.Py_DecRef` from thread 1
-#                 old_sticky = current_task().sticky
-#                 if !old_sticky
-#                     current_task().sticky = true
-#                 end
-#                 if Threads.threadid() == 1
-#                     for ptr in ptrs
-#                         if ptr != C.PyNULL
-#                             C.Py_DecRef(ptr)
-#                         end
-#                     end
-#                     handled = true
-#                 end
-#                 if !old_sticky
-#                     current_task().sticky = old_sticky
-#                 end
-#             end
-#             if !handled
-#                 for ptr in ptrs
-#                     if ptr != C.PyNULL
-#                         put!(QUEUE, ptr)
-#                     end
-#                 end
-#             end
-#         end
-#     end
-#     Threads.atomic_add!(SECONDS_SPENT_IN_GC, t)
-#     return
-# end
 
 # must only be called from thread 1 by the task in `GC_TASK[]`
 function unsafe_process_queue!()
--- a/src/GC/GC.jl
+++ b/src/GC/GC.jl
@@ -9,8 +9,19 @@ module GC
 
 using ..C: C
 
-const ENABLED = Ref(true)
-const QUEUE = C.PyPtr[]
+# `ENABLED`: whether or not python GC is enabled, or paused to process later
+const ENABLED = Threads.Atomic{Bool}(true)
+# this event allows us to `wait` in a task until GC is re-enabled
+const ENABLED_EVENT = Threads.Event()
+
+# this is the queue to process pointers for GC (`C.Py_DecRef`)
+const QUEUE = Channel{C.PyPtr}(Inf)
+
+# this is the task which performs GC from thread 1
+const GC_TASK = Ref{Task}()
+
+# This we use in testing to know when our GC is running
+const GC_FINISHED = Threads.Condition(ReentrantLock())
 
 """
     PythonCall.GC.disable()
@@ -24,6 +35,7 @@ Like most PythonCall functions, you must only call this from the main thread.
 """
 function disable()
     ENABLED[] = false
+    reset(ENABLED_EVENT)
     return
 end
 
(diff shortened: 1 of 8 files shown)


src/GC/GC.jl
```diff
@@ -9,77 +9,112 @@ module GC
...
-
-This frees any Python objects which were finalized while the GC was disabled, and allows
-objects finalized in the future to be freed immediately.
+Re-enable the PythonCall garbage collector. This should generally not be required.
 
-Like most PythonCall functions, you must only call this from the main thread.
 """
 function enable()
     ENABLED[] = true
+    notify(ENABLED_EVENT)
```
ericphanson, line 46: We could wait for the GC_finished event here to preserve the semantics that this function call only returns after all GC events are processed

cjdoris:
Very cool thanks. Will take me a while to absorb - I've not used Julia's multithreading much.

I'd be very interested to see some benchmarking to see how this affects performance in the single threaded case.

Another worry is what happens if we're running from JuliaCall? An issue with JuliaCall is that Julia's event loop doesn't run while in Python-land (unless you call `yield()` explicitly from Python), so if you do a bunch of stuff at the Python REPL, the PythonCall GC task might not run, maybe? I'd like to check.

ericphanson:
Do you have any good examples for a single-threaded python-object-heavy workload to benchmark? I think it's something that would benefit from a slightly realistic workload over just calling `py_finalizer` a bunch myself.

I also haven't used juliacall before so I'll have to figure out how to get that setup.

src/GC/GC.jl
```diff
@@ -9,77 +9,116 @@ module GC
...
+function launch_gc_task()
+    if isassigned(GC_TASK) && Base.istaskstarted(GC_TASK[]) && !Base.istaskdone(GC_TASK[])
+        throw(ConcurrencyViolationError("PythonCall GC task already running!"))
     end
-    return
+    task = Task(gc_loop)
+    task.sticky = VERSION >= v"1.7" # disallow task migration which was introduced in 1.7
+    # ensure the task runs from thread 1
+    ccall(:jl_set_task_tid, Cvoid, (Any, Cint), task, 0)
+    schedule(task)
```
ericphanson, line 110: this is from https://github.com/tro3/ThreadPools.jl/blob/44730372b9b00ba38192e5d2850dfe24c20cdbeb/src/macros.jl#L269-L276

ericphanson:
in the `gcbench.jl` script committed, I get:
```julia
Runtime: 15.873462 seconds (164.15 M allocations: 3.690 GiB, 24.07% gc time)
Python GC time: 2.328511 seconds
Python GC time (waiting): 0.000013 seconds
Next full GC: 2.217151 seconds (9.03% gc time)
Total: 18.104279 seconds (164.15 M allocations: 3.690 GiB, 22.21% gc time)
```
with Julia 1.10, with `-t1` to force single-threaded. I had to special case the case we are already on thread 1 to eagerly finalize the pointer instead of deferring to the queue, as the existing implementation does. This uses some distasteful manipulation of the current task's sticky bit to ensure it does not migrate while we are busy finalizing.

This matches the performance I get on `master`:
```julia
Runtime: 16.901158 seconds (164.15 M allocations: 3.690 GiB, 24.29% gc time)
Python GC time: 2.633090 seconds
Next full GC: 2.115611 seconds (9.28% gc time)
Total: 19.023249 seconds (164.15 M allocations: 3.690 GiB, 22.62% gc time)
```
(obtained with [these changes + script](https://github.com/JuliaPy/PythonCall.jl/compare/main...ericphanson:PythonCall.jl:eph/timing)).

The segfault in CI is concerning, it is on the pyjulia side which I have not looked into yet.

ericphanson:
> An issue with JuliaCall is that Julia's event loop doesn't run while in Python-land (unless you call yield() explicitly from Python), so if you do a bunch of stuff at the Python REPL, the PythonCall GC task might not run, maybe? I'd like to check.

’ve added some off-by-default logging so we can see when this background task kicks in. I’ll try running from pyjulia and see if it seems to be behaving as expected.

Btw do you mean the libuv event loop? I think the task runtime is separate from that but also I’m not really sure how it all works under the hood.

ericphanson:
I added a test for pyjulia and addressed https://github.com/JuliaPy/PythonCall.jl/issues/219#issuecomment-1712022932 by turning on signal handling if Julia is multithreaded, since threads + no signal handling is unsupported in Julia and leads to crashes. I also added threads to the CI matrix to prevent regressions on this.

cjdoris:
Thanks for your work so far.

Would be cool to implement and test the method from https://github.com/JuliaPy/PyCall.jl/pull/883 too - the simplicity (and avoidance of anything async) is appealing.

> Btw do you mean the libuv event loop?
No I meant the task scheduler, or whatever Julia calls the mechanism to run tasks asynchronously. My point was that in JuliaCall you might return to Python before the PythonCall GC task can do an iteration, and depending on what you're calling from Python it might never iterate.

MilesCranmer:
Looks awesome!

Regarding testing, it could be nice to try something like [rr's chaos mode](https://robert.ocallahan.org/2016/02/introducing-rr-chaos-mode.html) to assault the garbage collector and reveal any potential thread safety issues.

However, it is tricky to start Julia with rr from within Python – see https://github.com/JuliaLang/julia/issues/52689. So the test would have to be run from the Julia side of PythonCall.jl, and open Python from there, rather than the other direction. But maybe that would be good enough for testing this.

MilesCranmer:
Btw, @lassepe reports that this PR fixes the issues we had ran into here: https://github.com/avik-pal/Wandb.jl/issues/27

ericphanson:
> No I meant the task scheduler, or whatever Julia calls the mechanism to run tasks asynchronously. My point was that in JuliaCall you might return to Python before the PythonCall GC task can do an iteration, and depending on what you're calling from Python it might never iterate.

I see what you mean, if I stick
```julia
    t = Threads.@spawn begin
        while true
            println("Background task running $(time())")
            sleep(1)
        end
    end
    Base.errormonitor(t)
```

into the `__init__` method, then open python, import juliacall, I see the printout during loading of `juliacall`, but nothing when just sitting in the python repl. Calling methods like `rand(100)` etc via julia does not yield the printing, but if I call `sleep(10)` then I see it. So I think you're right, that the queue can go unprocessed when we aren't calling into julia code.

ericphanson:
BTW I wonder if this is an issue with the status quo solution as well? If a finalizer doesn't run (e.g. since GC hasn't run yet), the refcount won't be decremented. Do GC/finalizers run "in the background" properly in a way the tasks don't?

ericphanson:
I think another solution here would be to acquire the GIL from non-thread-1 threads (with [PyGILState_Ensure()](https://docs.python.org/3/c-api/init.html#c.PyGILState_Ensure) and release) and do the decrement eagerly within the finalizer. I'm not sure how performant that would be but it might be worth a try. The other thing is we want to make sure we can't deadlock somehow. Do we currently hold the GIL the whole time from thread 1?

cjdoris:
> I think another solution here would be to acquire the GIL from non-thread-1 threads (with [PyGILState_Ensure()](https://docs.python.org/3/c-api/init.html#c.PyGILState_Ensure) and release) and do the decrement eagerly within the finalizer. I'm not sure how performant that would be but it might be worth a try. The other thing is we want to make sure we can't deadlock somehow. Do we currently hold the GIL the whole time from thread 1?

Basically yes - thread 1 holds the GIL all the time.

Using PythonCall from Julia, the GIL is acquired implicitly when the Python runtime is initialised and released when it is exited. It is never released anywhere else by PythonCall.

Using JuliaCall from Python, the GIL is managed by Python as usual. Most of the time it is already acquired by the Python thread calling into PythonCall. Again, PythonCall never explicitly handles the GIL. The only exception to this is while PythonCall is being loaded, because this happens in a Julia C function being called from `ctypes` and this released the GIL, so we need to re-acquire it. Having read the docs just now, I could actually remove all of this and just use [`PyDLL`](https://docs.python.org/3/library/ctypes.html#ctypes.PyDLL) on the Python side to keep the GIL.

(Of course, you could call some Python code - such as ctypes or multithreading - which itself releases the GIL. But then it's currently documented that you should only call back into PythonCall from thread 1. Hence the above statement about the GIL always being held on thread 1 is still true while in any PythonCall code. If we're careful about it, we may be able to relax this.)

cjdoris:
> BTW I wonder if this is an issue with the status quo solution as well? If a finalizer doesn't run (e.g. since GC hasn't run yet), the refcount won't be decremented. Do GC/finalizers run "in the background" properly in a way the tasks don't?

Yep that's true. However with the status quo, the Python objects do get freed when Julia's GC is run. With the new solution, they only get freed after GC is run AND the async task you added gets yielded to. GC gets called frequently (it's hard to completely avoid allocating when using PythonCall) but most things you might do in JuliaCall probably don't yield.

cjdoris:
> > No I meant the task scheduler, or whatever Julia calls the mechanism to run tasks asynchronously. My point was that in JuliaCall you might return to Python before the PythonCall GC task can do an iteration, and depending on what you're calling from Python it might never iterate.
> 
> I see what you mean, if I stick
> 
> ```julia
>     t = Threads.@spawn begin
>         while true
>             println("Background task running $(time())")
>             sleep(1)
>         end
>     end
>     Base.errormonitor(t)
> ```
> 
> into the `__init__` method, then open python, import juliacall, I see the printout during loading of `juliacall`, but nothing when just sitting in the python repl. Calling methods like `m.rand(100)` etc via julia does not yield the printing, but if I call `m.sleep(10)` then I see it (for a julia module `m`). So I think you're right, that the queue can go unprocessed when we aren't calling into julia code.

If you call `m.Base.Libc.systemsleep(10)` instead, I assume you do not see the printing either? Just checking my understanding, because this should not yield.

cjdoris:
Some alternative approaches:
* In the finalizer: if on thread 1 then decref immediately and also decref anything in the queue; if on another thread then add to the queue. No async tasks here. This frees things quickly so long as there are Python objects being GC'd on thread 1 frequently.
* In the finaliser: if on thread 1 then decref immediately; if on another thread then add the finalizer again. This frees things quickly so long as the GC is called at all on thread 1 frequently (less restrictive than the previous option because the Python object remains un-finalized until GC runs on thread 1). Possibly slower, just because the finalizer might be called multiple times, not sure.
* This idea is kinda perverse but - instead of using an async task, we could create a mutable struct with a finalizer which just decrefs anything in the queue if on thread 1, and then adds the finalizer again to itself. This finalizer will be run periodically by the GC. Then Python objects being finalized just add their pointer to the queue (or maybe decref immediately if on thread 1).

cjdoris:
See #529 for a draft of an alternative approach without tasks.

cjdoris:
See #534 for a third approach to this. I think it's my favourite approach.

cjdoris:
Closing in favour of #529

PR #523, opened 2024-07-14
Require a compatible openssl version

### LINKED ISSUE #519: Libraries loaded via _jll packages are visible inside PythonCall environment.
Disclaimer: I am not sure whether this issue should be posted here or in `CondaPkg.jl`.

Related issues:
https://github.com/JuliaPackaging/Yggdrasil/issues/9028
https://github.com/avik-pal/Wandb.jl/issues/39

**Affects:** PythonCall 

**Describe the bug**
Over at Wandb.jl we have observed that, when `OpenSSL_jll@v3` is loaded before `Wandb.jl` (which wraps the Wandb python client via PythonCall), the package fails. When `OpenSSL_jll@v3` is loaded after `Wandb.jl` or when we downgrade to `OpenSSL_jll@v1`, everything works as expected. This suggests that the libraries loaded via julia _jll's propagate into the python environment. 

**Expected Behavior:**

I would hope that those two environments are isolated. If that is not possible, package authors using PythonCall should be made aware of having to pin JLL's at a version consistent with their dependencies (which is annoying but better than "hoping" that the right dependency gets loaded).

**Reproducer**

```julia
using OpenSSL_jll
using Wandb: Wandb

# trigger ssl error
function provoke_error(logger)
    logging_dict = Dict{String,Any}()
    logging_dict["samples"] = Wandb.Image(rand(2, 2))
    Wandb.log(logger, logging_dict; step=1)
end

logger = Wandb.WandbLogger(; project="test")
provoke_error(logger)
```

**System Info**
- Ubuntu 22.04
- Julia 1.10.4
- PythonCall 0.9.20

lassepe:
Resolves #519 (at least for the specific case of OpenSSL)
Requires https://github.com/JuliaPy/CondaPkg.jl/pull/140

Similar to how ibstdcpp-ng is handled, this PR communicated to `CondaPkg.jl` that we need a version of openssl that is compatible with one version potentially loaded via `OpenSSL_jll`

### DIFF
--- a/CondaPkg.toml
+++ b/CondaPkg.toml
@@ -1,6 +1,9 @@
 [deps.libstdcxx-ng]
 version = "<=julia"
 
+[deps.openssl]
+version = "<=julia"
+
 [deps.python]
 build = "**cpython**"
 version = ">=3.8,<4"
--- a/Project.toml
+++ b/Project.toml
@@ -18,7 +18,7 @@ UnsafePointers = "e17b2a0c-0bdf-430a-bd0c-3a23cae4ff39"
 
 [compat]
 Aqua = "0 - 999"
-CondaPkg = "0.2.21"
+CondaPkg = "0.2.23"
 Dates = "1"
 Libdl = "1"
 MacroTools = "0.5"


lassepe:
Note that the tests won't pass until a new version of `CondaPkg.jl` is tagged.

cjdoris:
Thanks :)

PR #529, opened 2024-07-24
More thread-safe GC

cjdoris:
A draft of an alternative to #520 to ensure Python GC only runs on thread 1.

This version avoids using tasks. Instead, when a Python object is finalized:
* If GC is enabled and the GIL is currently held, decref the object immediately.
* Otherwise add it to the queue.

I also added `PythonCall.GC.gc()` to explicitly clear the queue (if possible).

Furthermore I added an immortal object `GCHook()` which calls `PythonCall.GC.gc()` whenever it is finalized, then adds its own finalizer back on. This means the queue can be cleared periodically when Julia GC runs, because it keeps trying to finalize `GCHook()`.

I used `PyGILState_Check() == 1` to check that the GIL is currently held. This should make it safe to interact with Python from any thread in the future, provided you use the C API `PyGILState_*` and `PyThreadState_*` functions correctly.

Very much a draft. Need to use `Channel`s etc (like #520) to make this totally thread-safe.


ericphanson:
BTW, could be worth trying this benchmark: https://github.com/JuliaPy/PythonCall.jl/compare/main...ericphanson:PythonCall.jl:eph/timing

cjdoris:
@ericphanson @lassepe @MilesCranmer I think this is the version I'm going to merge.

I'd be grateful if you could give it a try to check it solves any issues, and also review the code.

Some timings on my machine:
* `pyint(123456789)` (quite a fast operation) takes 45ns on the main branch versus 54ns on this branch if the GIL is held.
* That slow-down is expected, because the main difference is the addition of a `C.PyGILState_Check()` call, which is 11ns.
* If the GIL is not held when the finalizer is called then the finalizer takes an extra 20ns, half being from locking/unlocking the spin lock, the other half being pushing/popping the queue. On main you get a slowdown of 10ns if GC is disabled (from pushing/popping), so this PR adds 10ns (locking/unlocking) in this situation.
* Using channels instead of vectors and locks adds about 80ns to the no-GIL branch. So I don't use channels.
* The fastest you can possibly create and free this int (`C.Py_DecRef(C.PyLong_FromLongLong(123456789))`) takes 19ns.
* In short, we were 2.25x slower than optimal before and now we are 2.75x slower. Or another way, we could make 22.2M ints per second before, now we can make 18.5M. This seems acceptable to me for the improvement.

I just need to copy over the extra tests from #520.

MilesCranmer:
PySR test suite is passing :+1:

MilesCranmer:
Cool, this also fixes https://github.com/avik-pal/Wandb.jl/issues/27!

src/GC/GC.jl
```diff
@@ -3,77 +3,135 @@
...
-See `disable` and `enable`.
+See [`enable`](@ref), [`disable`](@ref) and [`gc`](@ref).
 """
 module GC
 
 using ..C: C
 
-const ENABLED = Ref(true)
 const QUEUE = C.PyPtr[]
+const QUEUE_LOCK = Threads.SpinLock()
```
MilesCranmer, line 13: Nitpick is that I would prefer a pattern where these are one object so that it's more obvious to another developer that there is a lock associated to this queue, like:

```julia
const QUEUE = (;
    x=C.PyPtr[],
    lock=Threads.SpinLock()
)
```

and then you access it with

```julia
lock(QUEUE.lock) do
    f(QUEUE.x)
end
```

To me this is a more obvious association than the name + `_LOCK`. If a future dev does a global rename of `QUEUE`, they might forget to update the associated lock's name, and they would dissociate.

src/GC/GC.jl
```diff
@@ -3,77 +3,135 @@
...
         end
     end
-    return
+    nothing
+end
+
+"""
+    GCHook()
+
+An immortal object which frees any pending Python objects when Julia's GC runs.
```
MilesCranmer, line 109: Just wondering how/if this interacts with Julia multiprocessing and Python multiprocessing, if at all?

MilesCranmer (commented):
I think the docs also need to be updated with the warning about multithreading (and suggestion that `GC` be turned off).

Also, how does this interact with the signal handling, does the warning in the docs still apply?

src/GC/GC.jl
```diff
@@ -3,77 +3,135 @@
...
 
-Like most PythonCall functions, you must only call this from the main thread.
+    Historically this would disable the PythonCall garbage collector. This was required
+    for safety in multi-threaded code but is no longer needed, so this is now a no-op.
 """
-function disable()
-    ENABLED[] = false
-    return
-end
+disable() = nothing
```
MilesCranmer, line 26:
```suggestion
disable() = (Base.depwarn("disabling the PythonCall GC is no longer needed", :disable); nothing)
```

src/GC/GC.jl
```diff
@@ -3,77 +3,135 @@
...
+    for safety in multi-threaded code but is no longer needed, so this is now a no-op.
 """
-function enable()
-    ENABLED[] = true
-    if !isempty(QUEUE)
-        for ptr in QUEUE
-            if ptr != C.PyNULL
-                C.Py_DecRef(ptr)
-            end
+enable() = nothing
```
MilesCranmer, line 38:
```suggestion
enable() = (Base.depwarn("disabling the PythonCall GC is no longer needed", :enable); nothing)
```

src/GC/GC.jl
```diff
@@ -3,77 +3,135 @@
...
         end
     end
-    return
+    nothing
 end
 
 function enqueue_all(ptrs)
-    if C.CTX.is_initialized
-        if ENABLED[]
+    if any(ptr -> ptr != C.PYNULL, ptrs) && C.CTX.is_initialized
```
MilesCranmer, line 87:
```suggestion
    if any(Base.Fix2(!=, C.PYNULL), ptrs) && C.CTX.is_initialized
```
(reduces compilation load)

src/GC/GC.jl
```diff
@@ -3,77 +3,135 @@
...
+"""
+function gc()
+    if C.CTX.is_initialized
+        unsafe_free_queue()
+    end
+    nothing
+end
+
+function unsafe_free_queue()
+    lock(QUEUE_LOCK)
```
MilesCranmer, line 59: I prefer the `lock(...) do ...` API since it will still unlock in the case of an error.

src/GC/GC.jl
```diff
@@ -3,77 +3,135 @@
...
             for ptr in ptrs
                 if ptr != C.PyNULL
                     C.Py_DecRef(ptr)
                 end
             end
+            if !isempty(QUEUE)
+                unsafe_free_queue()
+            end
         else
+            lock(QUEUE_LOCK)
```
MilesCranmer, line 98: Prefer `lock(...) do ...`

src/GC/GC.jl
```diff
@@ -3,77 +3,135 @@
...
+thread currently holding the Python GIL.)
+"""
+function gc()
+    if C.CTX.is_initialized
+        unsafe_free_queue()
+    end
+    nothing
+end
+
+function unsafe_free_queue()
```
MilesCranmer, line 58: Why is it `unsafe_` if it is behind the lock? I would have assumed the `unsafe_` code would need to be locked from outside.

src/GC/GC.jl
```diff
@@ -3,77 +3,135 @@
 
 Garbage collection of Python objects.
 
-See `disable` and `enable`.
+See [`enable`](@ref), [`disable`](@ref) and [`gc`](@ref).
```
MilesCranmer, line 6: I'm assuming these can be removed from the module docstring since no longer in use?

src/GC/GC.jl
```diff
@@ -3,77 +3,135 @@
...
+    GCHook()
+
+An immortal object which frees any pending Python objects when Julia's GC runs.
+
+This works by creating it but not holding any strong reference to it, so it is eligible
+to be finalized by Julia's GC. The finalizer empties the PythonCall GC queue if
+possible. The finalizer also re-attaches itself, so the object does not actually get
+collected and so the finalizer will run again at next GC.
+"""
+mutable struct GCHook
```
MilesCranmer, line 116: Just curious, why `mutable`? (Maybe there's a deeper reason for interacting with Julia GC?)

src/GC/GC.jl
cjdoris: `Base.depwarn` is not documented and there doesn't seem to be an equivalent in the API

src/GC/GC.jl
cjdoris: As above, `Base.depwarn` is not API sadly

src/GC/GC.jl
cjdoris: Yeah I considered it but (a) it adds 30ns and (b) the code in between should not throw.

src/GC/GC.jl
cjdoris: See `gc()` - you need to check that `C.CTX.is_initialized` is true.

cjdoris:
> I think the docs also need to be updated with the warning about multithreading (and suggestion that `GC` be turned off).
> 
> Also, how does this interact with the signal handling, does the warning in the docs still apply?

Thanks I have updated the docs.

The signal handling warning still applies. #520 enabled it by default when multi-threading is enabled but I'm not changing it in this PR. Happy to discuss it though.

cjdoris:
I have copied the tests across from #520 and they're passing so I think this is ready to merge.

Will give you all a chance to comment.

cjdoris:
BTW this PR makes it possible to write code that calls into Python from multiple threads (or Python code calling into Julia from multiple threads), provided you handle the GIL correctly. I'm planning to add an API for safely acquiring and pausing the GIL after this PR.

src/GC/GC.jl
```diff
@@ -3,77 +3,134 @@
...
 
-Like most PythonCall functions, you must only call this from the main thread.
+    Historically this would disable the PythonCall garbage collector. This was required
+    for safety in multi-threaded code but is no longer needed, so this is now a no-op.
 """
-function disable()
-    ENABLED[] = false
-    return
-end
+disable() = nothing
```
MilesCranmer, line 25:
```suggestion
@deprecate disable() identity(nothing) false
```

This would be the documented equivalent. But normally if there's no alternative function to call you just write a message manually with `Base.depwarn`: https://github.com/search?q=/Base.depwarn/%20language:julia&type=code

src/GC/GC.jl
```diff
@@ -3,77 +3,134 @@
...
+    for safety in multi-threaded code but is no longer needed, so this is now a no-op.
 """
-function enable()
-    ENABLED[] = true
-    if !isempty(QUEUE)
-        for ptr in QUEUE
-            if ptr != C.PyNULL
-                C.Py_DecRef(ptr)
-            end
+enable() = nothing
```
MilesCranmer, line 37:
```suggestion
@deprecate enable() identity(nothing) false
```

src/GC/GC.jl
MilesCranmer: (It's helpful for library maintainers, since the `depwarn` will show up during unittests but not during normal execution – so I know what functions I need to update.)

MilesCranmer: Since if you do it with the `@deprecated` API, it doesn't really make sense in this context:

```julia
julia> function foo()
       end
foo (generic function with 1 method)

julia> function _foo()
       end
_foo (generic function with 1 method)

julia> @deprecate foo() _foo()
foo (generic function with 1 method)

julia> foo()
┌ Warning: `foo()` is deprecated, use `_foo()` instead.
│   caller = top-level scope at REPL[4]:1
└ @ Core REPL[4]:1
```

src/GC/GC.jl
MilesCranmer: Though again, I'd prefer the `Base.depwarn`, otherwise it's a bit confusing to package maintainers

src/GC/GC.jl
MilesCranmer: I guess it's the additional function call from the closure. In that case I would try

```julia
lock(QUEUE_LOCK)
try
    #= do stuff =#
finally
    unlock(QUEUE_LOCK)
end
```
And hopefully that's a bit faster.

If anything it's just a good habit. I think there's always a possibility for some kind of error in Julia code until we can actually impose such constraints on code, so locks should always be safely handled. e.g., perhaps the user could `<Ctrl-C>` while the lock is in place, or there's an OOM error, or something else we haven't thought of.

src/GC/GC.jl
```diff
@@ -3,77 +3,135 @@
...
-        if ENABLED[]
+        if C.PyGILState_Check() == 1
             C.Py_DecRef(ptr)
+            if !isempty(QUEUE)
+                unsafe_free_queue()
+            end
         else
+            lock(QUEUE_LOCK)
             push!(QUEUE, ptr)
+            unlock(QUEUE_LOCK)
```
MilesCranmer:
```suggestion
            lock(QUEUE_LOCK)
            try
                push!(QUEUE, ptr)
            finally
                unlock(QUEUE_LOCK)
            end
```

src/GC/GC.jl
MilesCranmer: (As above)

MilesCranmer:
Regarding `Base.depwarn`, it looks like it was just officially documented and declared part of the public API: https://github.com/JuliaLang/julia/blame/19165bec4a3e2ed0de4f2f43bccc3fcac2d76662/base/deprecated.jl#L230-L249

src/GC/GC.jl
ericphanson: BTW `Base.@lock` (which is what I used in my PR) expands to exactly the code you wrote there @MilesCranmer, and is exported on more recent julia versions

src/GC/GC.jl
```diff
@@ -3,77 +3,134 @@
...
-See `disable` and `enable`.
+See [`gc`](@ref).
 """
 module GC
 
 using ..C: C
 
-const ENABLED = Ref(true)
-const QUEUE = C.PyPtr[]
+const QUEUE = (; items = C.PyPtr[], lock = Threads.SpinLock())
```
ericphanson, line 12: Base has `Lockable` starting in 1.11 (exported in 1.12) FYI. So not useful here but just so you know :). It is just like this, except it also enforces the lock is held when you access the item it holds via indexing.

src/GC/GC.jl
MilesCranmer: Note this is also the same pattern recommended by the docs: https://docs.julialang.org/en/v1/base/parallel/#Base.ReentrantLock as well as the pattern that is generated by the `Base.@lock` macro: https://docs.julialang.org/en/v1.11.0-rc2/base/parallel/#Base.@lock

src/GC/GC.jl
MilesCranmer: Unresolving this

src/GC/GC.jl
MilesCranmer, line 25:
```suggestion
disable() = (Base.depwarn("disabling the PythonCall GC is no longer needed", :disable); nothing)
```
Unresolving this as we've identified the docs and this function being official public API: https://github.com/JuliaLang/julia/pull/55212

src/GC/GC.jl
MilesCranmer, line 37:
```suggestion
enable() = (Base.depwarn("disabling the PythonCall GC is no longer needed", :disable); nothing)
```

Unresolving this as we've identified the docs and this function being official public API: https://github.com/JuliaLang/julia/pull/55212

MilesCranmer:
I think apart from putting the lock behind a `try`-`finally`, it's otherwise looking good.

My one other comment is more of a question (for both you and @ericphanson) – why `Threads.SpinLock()` and not `ReentrantLock()`?

ericphanson:
I think it is because we can't yield to other tasks in finalizers so we have to busywait: 

> f must not cause a task switch, which excludes most I/O operations such as println. Using the @async macro (to defer context switching to outside of the finalizer) or ccall to directly invoke IO functions in C may be helpful for debugging purposes.


from https://docs.julialang.org/en/v1/base/base/#Base.finalizer

MilesCranmer:
I think I'm misreading, because a reentrant lock wouldn't allow that, right? Reentrant locks are about letting the same task acquire the lock multiple times?

src/GC/GC.jl
```diff
@@ -3,77 +3,134 @@
...
         else
-            append!(QUEUE, ptrs)
+            lock(QUEUE.lock)
+            append!(QUEUE.items, ptrs)
+            unlock(QUEUE.lock)
         end
     end
-    return
+    nothing
+end
```
MilesCranmer, line 103: Tangentially related, could each of these functions have a few more inline comments? Just to help me figure out the purpose of each block.

ericphanson:
I edited in a response a bit late sorry. ReentrantLock vs SpinLock is not only about reentry but also implementation, since task-switching is an implementation detail. ReentrantLocks do it and SpinLocks don't.

See: [ReentrantLock waiting](https://github.com/JuliaLang/julia/blob/19165bec4a3e2ed0de4f2f43bccc3fcac2d76662/base/lock.jl#L157)

[SpinLock implementation](https://github.com/JuliaLang/julia/blob/19165bec4a3e2ed0de4f2f43bccc3fcac2d76662/base/locks-mt.jl#L41-L71)

src/GC/GC.jl
ericphanson: It won't let me suggest comments on deleted lines, but my attempt:

* `any(!=(C.PYNULL), ptrs)`: is there any work to do? if the pointers are all null then nothing to do
* `C.CTX.is_initialized`: need to ensure the context is initialized (not sure what that does?)
* `C.PyGILState_Check() == 1`: does this thread hold the GIL?
* if so, eagerly decrease the reference count on each object without putting it in the queue. Then since we know we hold the GIL, for good measure also process the queue built up on other threads
* if not, then acquire the queue lock so we can add these items to the queue

MilesCranmer:
Maybe if we use the same pattern from https://github.com/JuliaPy/PyCall.jl/pull/1074 we can get a speedup? If we have a `lock(::SpinLock)`, the GC threads are forced to wait for each other. But if we do the equivalent of the following code from PyCall (never implemented; just a PR)

```julia
function _pydecref_locked(po::PyObject)
    # If available, we lock and decref
    !islocked(PYDECREF_LOCK) &&
        trylock(() -> pydecref_(po), PYDECREF_LOCK) &&
        return nothing

    # Otherwise, add back to queue to be decref'd later
    finalizer(_pydecref_locked, po)
    return nothing
end
```

then the threads never need to have a busy wait on the lock to be released, they just add it back to the queue and get on with some other work.

src/GC/GC.jl
```diff
@@ -3,77 +3,134 @@
...
+        finalizer(_gchook_finalizer, x)
+        if !isempty(QUEUE.items) && C.PyGILState_Check() == 1
+            unsafe_free_queue()
+        end
+    end
+    nothing
+end
+
+function __init__()
+    HOOK.value = GCHook()
```
ericphanson, line 132: hm, I'm not sure this does what you want. The `WeakRef` constructor does a [ccall](https://github.com/JuliaLang/julia/blob/19165bec4a3e2ed0de4f2f43bccc3fcac2d76662/base/boot.jl#L451) and we are bypassing that.

What about `HOOK` being a Ref holding a weakref? 😅. Then we can construct the weakref via its constructor and stick it in `HOOK`.

ericphanson:
That is https://github.com/JuliaPy/PythonCall.jl/pull/534, right? maybe that strategy should be discussed in that PR

MilesCranmer:
Oh yeah good point:

> To me some kind of queue makes the most sense. However probably anything is better than the segfaults of the status quo.

I agree, I think we can just worry about further optimizations later.

src/GC/GC.jl
cjdoris: The way I did it did appear to work. However assigning to `.value` is not documented behaviour so I have switched to your suggestion. We're only keeping the weak reference around for debugging purposes.

ericphanson:
I think we need SpinLock since we are in a finalizer! We aren’t allowed to yield like ReentrantLock does

MilesCranmer (approved):
Looks great!

pytest/test_all.py
```diff
@@ -75,3 +75,28 @@ def test_issue_433():
         """
     )
     assert out == 25
+
+def test_julia_gc():
+    from juliacall import Main as jl
+    # We make a bunch of python objects with no reference to them,
+    # then call GC to try to finalize them.
+    # We want to make sure we don't segfault.
+    # Here we can (manually) verify that the background task is running successfully,
```
ericphanson, line 84: This comment is out of date (no background task, no printing)

ericphanson (approved):
One out of date comment, otherwise lgtm (though I should say I’m not expert and was just learning as I went in #520)

src/GC/GC.jl
```diff
@@ -3,77 +3,152 @@
...
+mutable struct GCHook
+    function GCHook()
+        finalizer(_gchook_finalizer, new())
+    end
+end
+
+function _gchook_finalizer(x)
+    if C.CTX.is_initialized
+        finalizer(_gchook_finalizer, x)
+        if !isempty(QUEUE.items) && C.PyGILState_Check() == 1
```
MilesCranmer, line 142: Does this `isempty` need to be locked?

Looking at the implementation: https://github.com/JuliaLang/julia/blob/48d4fd48430af58502699fdf3504b90589df3852/base/essentials.jl#L957 it might need to be.

src/GC/GC.jl
```diff
@@ -3,77 +3,152 @@
...
-    if C.CTX.is_initialized
-        if ENABLED[]
+    if any(!=(C.PYNULL), ptrs) && C.CTX.is_initialized
+        if C.PyGILState_Check() == 1
             for ptr in ptrs
                 if ptr != C.PyNULL
                     C.Py_DecRef(ptr)
                 end
             end
+            if !isempty(QUEUE.items)
```
MilesCranmer, line 113: i think the entire if statement should be locked, rather than just the call to `append!`. See my other comment — you might otherwise be trying to `iterate` through since-removed elements.

src/GC/GC.jl
```diff
@@ -3,77 +3,152 @@
...
+    # If the ptr is NULL there is nothing to free.
+    # If C.CTX.is_initialized is false then the Python interpreter hasn't started yet
+    # or has been finalized; either way attempting to free will cause an error.
     if ptr != C.PyNULL && C.CTX.is_initialized
-        if ENABLED[]
+        if C.PyGILState_Check() == 1
+            # If the current thread holds the GIL, then we can immediately free.
             C.Py_DecRef(ptr)
+            # We may as well also free any other enqueued objects.
+            if !isempty(QUEUE.items)
```
MilesCranmer, line 90: Similarly, the whole of statement should be locked.

MilesCranmer: Or at least the `isempty` call by itself. I think it can segfault if another thread is making concurrent writes

src/GC/GC.jl
cjdoris: No it's this method (just checking length == 0): https://github.com/JuliaLang/julia/blob/1dffd7752de2409b5f6c81fdcc1f33118127725e/base/abstractarray.jl#L1237

I believe this is safe. At worst, if there is a race condition, the `isempty` check might return `true` when it should be `false` or vice-versa, but all this means is that we might not empty the queue when we should, or we might waste time emptying an already-empty queue. But neither of these is really a problem - we'll just empty the queue next time around. It's suuuuuper unlikely such a race condition would strike many times in succession.

src/GC/GC.jl
cjdoris: All mutation of the queue is behind the lock, so I think this is fine given my other comment about `isempty`.

src/GC/GC.jl
cjdoris: (ditto my other comment)

MilesCranmer:
Regarding the `isempty` needing to be locked,
```julia
iterate(items) === nothing
```
This is how Base checks for an array being empty. I think what can happen is: thread A can (1) call the `iterate`, (2) get the pointer to the first array element, and then (3) a separate thread B removes that element from the array, making that pointer point to invalid memory, (4) thread A accesses the first array element anyways.

In other words, use of `isempty` could cause undefined behavior if another thread is writing to the collection (and maybe a segfault).

cjdoris:
> Regarding the `isempty` needing to be locked,

I [commented elsewhere](https://github.com/JuliaPy/PythonCall.jl/pull/529#discussion_r1702283046) that you're looking at the wrong method of `isempty`. [The method for arrays](https://github.com/JuliaLang/julia/blob/master/base/abstractarray.jl#L1237) is
```julia
isempty(a::AbstractArray) = (length(a) == 0)
```
which I believe is safe. It might give wrong answers occasionally in the presence of a race condition, but it's safe.

MilesCranmer:
Oh yeah, you are totally right, sorry.

MilesCranmer:
Ok, no other comments from me. I think it's ready to go :+1:

cjdoris:
Thanks for your help reviewing!

MilesCranmer:
No problem :+1:

(By the way, could we get a release so we can fix the Wandb.jl issue?)

cjdoris:
Yeah I'll make a release once #535 is merged (going to sleep on it and bikeshed the new function names).

MilesCranmer:
Sounds good.

In case #535 seems like it will take some more discussion to settle on an API I wonder if we could have a patch release just so we can get the bug fixes through?

MilesCranmer:
@cjdoris could we have a release before #535? I need the thread safety for something PySR-related and it's slightly urgent (am teaching a summer school this week and want to have them using Wandb.jl)

cjdoris:
It's released

MilesCranmer:
Thanks!

PR #535, opened 2024-08-02
GIL functions for genuine multi-threading

cjdoris:
Adds `PythonCall.GIL` module with:
- `lock(f)` to execute `f()` with the GIL held
- `@lock expr` the macro equivalent
- `release(f)` to execute `f()` with the GIL released
- `@release expr` the macro equivalent

Here's an example from the tests. It calls Python's `time.sleep(1)` twice in two Julia threads. `GIL.@release` is required to release the GIL so that within the threads `GIL.@lock` can re-acquire it. Since `time.sleep` itself then releases the GIL again, the sleep actually happens in parallel and the whole block takes 1 second.
```julia
PythonCall.GIL.@release Threads.@threads for i = 1:2
    PythonCall.GIL.@lock pyimport("time").sleep(1)
end
```

Here's a similar example from the Python tests. It defines a Julia function `jsleep` which behaves pretty much like `time.sleep`. In particular it also releases the GIL, so when you call `jsleep(1)` twice in parallel, it only takes 1 second in total.
```python
from concurrent.futures import ThreadPoolExecutor, wait
jsleep = jl.seval("t::Real -> PythonCall.GIL.@release Libc.systemsleep(t)")
pool = ThreadPoolExecutor(2)
fs = [pool.submit(jsleep, 1) for _ in range(2)]
wait(fs)
```

When using this from the Python side, if the Julia code can yield (e.g. `sleep(1)`), the threaded tasks won't finish unless you explicitly yield back from Python to let the yielding code finish. You'll probably need to yield in a loop. We could provide some functionality for this in JuliaCall but I'll leave that for another ticket.

- [x] release notes
- [x] julia tests
- [x] python tests
- [x] documentation - I think a lot of the advice in the FAQ could move into a new GIL/multithreading section in the main docs.
- [x] builds on #529 - merge that first


cjdoris:
@MilesCranmer @ericphanson thought you'd like this PR - I think it's cool

MilesCranmer:
This is awesome!!

MilesCranmer:
Yeah I think this would be very very useful for the PySR/SymbolicRegression.jl ecosystem. The most requested feature is being able to use Python functionality from within user-defined custom loss functions (which are run in parallel). I think this feature would help enable that.

cjdoris:
Ok I'm happy with this except the names `lock` and `release` which feel asymmetric. My suggestions in order of preference:
- `acquire` and `release` (Python uses this terminology in the C API docs, though it also uses ensure/release and restore/save)
- `lock` and `unlock` (more like the Julia stdlib)
- `pygil_acquire` and `pygil_release` and export these.

MilesCranmer:
I strongly prefer `lock` and `unlock`. Here's my reasoning:

Even though I use Python for 50% of my work, I have literally never seen `acquire` and `release` used before. I'm assuming this is because the Python multithreading/multiprocessing API is 99% of the time uses immutable storage and interacts with a mapped function on a thread pool. In other words, thread locks are fairly uncommon in Python, and I think most people interacting with this interface will be more used to the Julia versions.

In Julia, `lock` and `unlock` are fairly common functions. So much so that they are found in `Base`, which is unlike Python where `acquire` is only found in a separate stdlib modules (like [here](https://docs.python.org/3/library/threading.html#threading.RLock.acquire)).

Though these functions control the Python API, the exposed API is called via PythonCall.jl. Thus, I think they should respect the Julia standards. Perhaps a `juliacall` version could use `acquire` and `release`. Basically whenever there is an ambiguity in a syntax choice, I think PythonCall.jl should prefer normal Julia syntax, and juliacall should prefer normal Python syntax.

MilesCranmer:
One other thought: would it ever be possible to control multiple Python processes from PythonCall.jl? In this case I wonder if there should be some kind of `lock(f::Function, lock::PythonProcessLock)` implemented, so you can choose a specific Python process to lock the GIL of.

MilesCranmer:
Still though, it might be nice to have a Julian version, so you could really use the standard patterns (including other tools that might interface with `Base.lock`, like:

```julia
lock(PythonCall.GIL) do
    #= do stuff =#
end
```

Rather than `PythonCall.GIL.Lock`.

MilesCranmer:
Argh, but then again, I guess the GIL is kind of different than a normal Julia lock... So that might motivate having different names for these.

cjdoris:
Yeah I think I agree that `lock`, `unlock` is right. This is a Julia package so should use Julia terminology - which has been my approach elsewhere in PythonCall and CondaPkg.

It's a nice idea to try overloading Base functions for locking, but it's not immediately obvious how, so I'll leave that for a separate PR.

MilesCranmer:
> It's a nice idea to try overloading Base functions for locking, but it's not immediately obvious how, so I'll leave that for a separate PR.

The tricky thing here is that it would require a slightly different syntax. The `lock` in `Base` locks a specific object, whereas this one takes no arguments. This is kind of why I feel like `lock(PythonCall.GILRef)` might be nice. And `GILRef` can just be an abstract type to allow declaring a `Base.lock(::Type{PythonCall.GILRef})` 

(Not sure the right name here... `GILRef` doesn't sound great)

PR #538, opened 2024-08-15
Update GIL.jl lock's documentation (possible mistake?)

camilodlt:
I think "unlock" is a typo and should be "lock". 
Hence, a GIL lock "locks, computes, unlocks" rather than "unlocks, computes, unlocks" unless there is something I'm missing. 

Nice work on the GIL by the way !


cjdoris:
Well spotted, thanks!

PR #545, opened 2024-08-23
Switch to pyproject.toml

MilesCranmer:
`pyproject.toml` is the standard mechanism for a Python project's metadata while `setup.cfg` is a deprecated format: https://peps.python.org/pep-0518.

This PR simply translates everything from one format to the other.

Note that this is required to use some build systems locally, such as rye, which do not support the older formats.


MilesCranmer:
Friendly ping @cjdoris :)

cjdoris:
Great thanks! I'd been meaning to do this and also switch to the hatchling build backend for a while, so you've done 80% of that task :)

PR #552, opened 2024-09-06
use uv and hatchling

cjdoris:


MilesCranmer (commented):
This is looking good (and super useful for CI). Though I ran into the following bug when I ran the tests locally for some reason; is this expected?

```python
[ins] In [5]: from juliacall import Main as jl

[ins] In [6]: jl.x = 1
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[6], line 1
----> 1 jl.x = 1

File ~/.julia/packages/PythonCall/Nr75f/src/JlWrap/any.jl:254, in __setattr__(self, k, v)
    252     else:
    253         return
--> 254     self._jl_callmethod($(pyjl_methodnum(pyjlany_setattr)), k, v)
    255 def __dir__(self):
    256     return ValueBase.__dir__(self) + self._jl_callmethod($(pyjl_methodnum(pyjlany_dir)))

AttributeError: Julia: Global Main.x does not exist and cannot be assigned. Declare it using `global` before attempting assignment.
```

Issue #582, opened 2025-01-01
Julia 1.11 prevents assigning objects to Main


MilesCranmer:
It looks like this syntax is no longer working on Julia 1.11:

```python
from juliacall import Main as jl

jl.x = 1
```

On 1.10 of Julia this works, but on 1.11, this raises the following error:

```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
    import juliapkg as jp; jp.require_julia("~1.11"); from juliacall import Main as jl; jl.x = 1
                                                                                        ^^^^
  File "/Users/mcranmer/PermaDocuments/PythonCall.jl/src/JlWrap/any.jl", line 254, in __setattr__
    self._jl_callmethod($(pyjl_methodnum(pyjlany_setattr)), k, v)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: Julia: Global Main.x does not exist and cannot be assigned. Declare it using `global` before attempting assignment.
```

I also verified this error did not depend on Python (3.10 through 3.13 all give the same error).

To quickly verify this:

```bash
# 1.10
python -c 'import juliapkg as jp; jp.require_julia("~1.10"); from juliacall import Main as jl; jl.x = 1'
# (Exits fine)

# 1.11
python -c 'import juliapkg as jp; jp.require_julia("~1.11"); from juliacall import Main as jl; jl.x = 1'
# Throws an error
```

cc @cjdoris

MilesCranmer:
x-post: https://github.com/JuliaLang/julia/issues/56933

PR #583, opened 2025-01-01
fix: patch changed behavior of `setproperty!` for modules

MilesCranmer:
This fixes the breaking change introduced by Julia 1.11: https://github.com/JuliaLang/julia/pull/54678. See issue #582 for discussion.

The fix is pretty simple - we just automatically call `global` on variables that are undefined. This means that the syntax

```python

jl.x = 1
```

will continue to function.

@cjdoris could you please take a look and merge? Since Julia 1.11 is out now I think we should get this fix in relatively quickly since a lot of the docs and tutorials rely on this behavior working.

### DIFF
--- a/src/JlWrap/any.jl
+++ b/src/JlWrap/any.jl
@@ -26,6 +26,10 @@ function pyjlany_setattr(self, k_::Py, v_::Py)
     k = Symbol(pyjl_attr_py2jl(pyconvert(String, k_)))
     pydel!(k_)
     v = pyconvert(Any, v_)
+    if VERSION >= v"1.11.0-" && self isa Module && !isdefined(self, k)
+        # Fix for https://github.com/JuliaLang/julia/pull/54678
+        Base.Core.eval(self, Expr(:global, k))
+    end
     setproperty!(self, k, v)
     Py(nothing)
 end


test/Aqua.jl
```diff
@@ -2,5 +2,9 @@
     # The unbound_args test fails on methods with signature like foo(::Type{Tuple{Vararg{V}}}) where V
```
MilesCranmer, line 2: Unrelated but this is a real unbound_args. `V` is unbound in `Tuple{Vararg{V}}` because `()` is technically a subtype - in which case `V` has no definition.

Correct signature with bounded `V` would be `Tuple{V,Vararg{V}}`

MilesCranmer:
FYI I'm ignoring some of the broken tests (only two; seem minor) temporarily since I see the `setproperty!` issue as being much more impactful

MilesCranmer:
The only remaining test error is from the julia_gc test. I think it's more likely this is due to Julia bugs since 1.11's parallel GC which has some internal issues (e.g., https://github.com/JuliaLang/julia/issues/56871, https://github.com/JuliaLang/julia/issues/56759, https://github.com/JuliaLang/julia/issues/56735).

I'll turn off that test for 1.11 too, since it doesn't identify an issue with PythonCall.jl (if that's alright with you). We can look into it later once 1.11.3 comes out.

MilesCranmer:
@cjdoris we're getting errors in other PRs that require this behavior; do you think this could be merged when you get a chance?

cjdoris:
Thanks - I spent some time understanding the various test changes and made some tweaks/fixes in the process. Will make a release.

MilesCranmer:
Awesome, thanks!

PR #588, opened 2025-01-19
PyList: fix pushfirst! for Julia 1.11, by adding prepend!

hhaensel:
Julia 1.11 handles pushfirst! with multiple args differently and fails if no prepend! is defined.
This PR defines the appropriate prepend! method and adds respective tests.


cjdoris (commented):
Thanks!

PR #589, opened 2025-01-20
fix pandas testing

hhaensel:


hhaensel:
Cherry-picked into hh-timedelta64

Issue #361, opened 2023-09-19
Better show for numpy array


xgdgsc:
**Is your feature request related to a problem? Please describe.**
https://github.com/JuliaPy/PythonCall.jl/blob/ad67c0457ce79f13deb0d04a47808824f8ad7921/src/Py.jl#L186 When it shows a python array in vscode inline results. It would only show `Python` when unexpanded because of the line end. Would it be better to show more informative like Julia Array shows `4×4 Matrix{Float64}:` on the first line?

**Describe the solution you'd like**
shows numpy array shape like `4×4 Matrix{Float64}:`

**Describe alternatives you've considered**
 

**Additional context**

cjdoris:
I'm hesitant to make this code depend on the actual Python type - currently this basically just shows `repr(x)`. I'm not totally against the idea though - I agree it's useful info.

Maybe you could suggest this change to numpy itself?

xgdgsc:
Would there still be a first line `Python: ` ? How feasible is it to add something like https://github.com/numpy/numpy/issues/9029#issuecomment-298242528 ?

cjdoris:
Oh I see your point now.

To explain for myself: when a `Py` is shown, if it has multiple lines then the `Python:` prefix occurs on its own line, such as with numpy arrays:
```
Python:
array(
  [ ... ], dtype=int64)
```
If you execute code line-by-line from a Julia script in VS Code then it only displays the first line (`"Python:"`) which is not very informative. So this question is about adding extra info into the first line.

PR #595, opened 2025-02-28
show more arr metadata inline

xgdgsc:
fix https://github.com/JuliaPy/PythonCall.jl/issues/361


xgdgsc:
@cjdoris any interest on reviewing?

cjdoris:
Yep the example was using your branch, but I made some modifications playing around with it to see what worked.

I think what actually might be a nice simple generic thing would be to have it show like
```
Py: array([[ 0,  1,  2,  3,  4], ...
```

This would only need two very small changes:
1. Don't insert a newline in the prefix when in compact mode (which is the right thing to do since compact is supposed to output a single line).
2. Change the VS Code extension to print this bit with `compact=true` (which makes sense anyway because it wants a single line anyway).

Thoughts?

PR #616, opened 2025-05-20
Allow pydict accept Pair args

mkitti:
Create a method `pydict(pairs::Pair...)` similar to `Dict(::Pair...)`.

This allow for the following usage.

```julia
julia> using PythonCall

julia> Dict("a" => 5, "b" => 6)
Dict{String, Int64} with 2 entries:
  "b" => 6
  "a" => 5

julia> pydict("a" => 5, "b" => 6)
Python: {'a': 5, 'b': 6}
```


mkitti:
@cjdoris There seems to be an issue with Python 3.13 per the failing test.

xref: https://github.com/JuliaLang/julia/issues/58475

cjdoris:
Good idea, thanks.

PR #618, opened 2025-05-28
use getptr a lot less

### LINKED ISSUE #617: Segmentation fault when calling basic python functions
**Affects:** PythonCall

**Describe the bug**
I get segmentation faults when using PythonCall to call python functions that create python objects many times. It does not happen deterministically, but happens (on a couple of systems I have tested it on) reliably for the MWE below:

```jl
using PythonCall
N = 100_000_000
for i in 1:N
    args = pytuple((2.0, 1.0, 0.0))
    @assert !pyeq(Bool, args[0], args[2]) # SEGFAULT
    if i % 100_000 == 0
        println("step $i ($(i/N*100) %)")
    end
end
```
Sometimes I get the segmentation fault immediately, before the first print statement, sometimes 30% of the loop succedes before the segfault.

If I wrap the doubles in `Py` objects explicitly before, I do not get segmentation faults:
```jl
using PythonCall
N = 100_000_000
for i in 1:N
    args = pytuple((Py(2.0), Py(1.0), Py(0.0)))
    @assert !pyeq(Bool, args[0], args[2]) # NO SEGFAULT
    if i % 100_000 == 0
        println("step $i ($(i/N*100) %)")
    end
end
```
My understanding is that this shouldn't be necessary to deal with explicitly, but somehow it is in this example. And since this is only a MWE it is not a solution to my actual problem, since I do not know which calls / arguments need this explicit handling and which do not.

Here is the segfault output (from the **first** code snippet):
```
[872756] signal 11 (1): Segmentation fault
in expression starting at /home/fischer/dev/jl/PythonCallSegFault/pythoncall_segfault_mwe.jl:5
do_richcompare at /usr/local/src/conda/python-3.12.10/Objects/object.c:815 [inlined]
PyObject_RichCompare at /usr/local/src/conda/python-3.12.10/Objects/object.c:865 [inlined]
PyObject_RichCompareBool at /usr/local/src/conda/python-3.12.10/Objects/object.c:887
PyObject_RichCompareBool at /home/fischer/.julia/packages/PythonCall/L4cjh/src/C/pointers.jl:303
unknown function (ip: 0x7efe9e9fca05)
unknown function (ip: 0x7efe9e9fc699)
unknown function (ip: 0x7efe9e9fc623)
macro expansion at /home/fischer/.julia/packages/PythonCall/L4cjh/src/Core/Py.jl:132 [inlined]
pyeq at /home/fischer/.julia/packages/PythonCall/L4cjh/src/Core/builtins.jl:304 [inlined]
top-level scope at /home/fischer/dev/jl/PythonCallSegFault/pythoncall_segfault_mwe.jl:12
jl_toplevel_eval_flex at /cache/build/tester-amdci5-12/julialang/julia-release-1-dot-11/src/toplevel.c:934
jl_toplevel_eval_flex at /cache/build/tester-amdci5-12/julialang/julia-release-1-dot-11/src/toplevel.c:886
ijl_toplevel_eval_in at /cache/build/tester-amdci5-12/julialang/julia-release-1-dot-11/src/toplevel.c:994
eval at ./boot.jl:430 [inlined]
include_string at ./loading.jl:2734
_include at ./loading.jl:2794
include at ./Base.jl:557
jfptr_include_46879.1 at /home/fischer/.julia/juliaup/julia-1.11.5+0.x64.linux.gnu/lib/julia/sys.so (unknown line)
exec_options at ./client.jl:323
_start at ./client.jl:531
jfptr__start_73430.1 at /home/fischer/.julia/juliaup/julia-1.11.5+0.x64.linux.gnu/lib/julia/sys.so (unknown line)
jl_apply at /cache/build/tester-amdci5-12/julialang/julia-release-1-dot-11/src/julia.h:2157 [inlined]
true_main at /cache/build/tester-amdci5-12/julialang/julia-release-1-dot-11/src/jlapi.c:900
jl_repl_entrypoint at /cache/build/tester-amdci5-12/julialang/julia-release-1-dot-11/src/jlapi.c:1059
main at /cache/build/tester-amdci5-12/julialang/julia-release-1-dot-11/cli/loader_exe.c:58
__libc_start_main at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
unknown function (ip: 0x4010b8)
Allocations: 312994585 (Pool: 312994469; Big: 116); GC: 13
[1]    872756 segmentation fault  julia pythoncall_segfault_mwe.jl
```

**Your system**
Please provide detailed information about your system:
- OS: Ubuntu 20.04
```jl
julia> versioninfo()
Julia Version 1.11.5
Commit 760b2e5b739 (2025-04-14 06:53 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × 12th Gen Intel(R) Core(TM) i7-1260P
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, alderlake)
Threads: 1 default, 0 interactive, 1 GC (on 16 virtual cores)

julia> Pkg.status()
Status `~/dev/jl/PythonCallSegFault/Project.toml`
  [992eb4ea] CondaPkg v0.2.29
  [6099a3de] PythonCall v0.9.25

julia> CondaPkg.status()
CondaPkg Status /home/fischer/dev/jl/PythonCallSegFault/CondaPkg.toml (empty)
Environment
  /home/fischer/dev/jl/PythonCallSegFault/.CondaPkg/.pixi/envs/default

julia> pyimport("sys").version
Python: '3.12.10 | packaged by conda-forge | (main, Apr 10 2025, 22:41:16) [GCC 13.3.0]'
```

**Additional context**
The actual context where this happened to me is that I am using PythonCall.jl to use the python interface of the [lanelet2](https://github.com/fzi-forschungszentrum-informatik/Lanelet2) library, which in turn is a boost python interface to C++ code. I am getting segmentation faults from a couple of calls to this library.

cjdoris:
Defines an `unsafe_convert` rule for `Py` to `C.PyPtr` so we can pass `Py` directly to C functions.

Also defines `incref(::Py)` so we don't have to operate on pointers.

Generalises our faked C functions in `extras.jl` to also allow `::Py` arguments in place of `::PyPtr`.

Fixes #617 - and more generally removes a lot of places where the Julia GC could have freed a `Py` while we still have a pointer to the Python object, causing a read-after-free error.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,7 @@ jobs:
     name: Test (${{ matrix.os }}, julia ${{ matrix.jlversion }})
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         arch: [x64] # x86 unsupported by MicroMamba
         os: [ubuntu-latest, windows-latest, macos-latest]
@@ -53,7 +53,7 @@ jobs:
     name: Test (${{ matrix.os }}, python ${{ matrix.pyversion }})
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         pyversion: [">=3.8", "3.8"]
--- a/src/C/extras.jl
+++ b/src/C/extras.jl
@@ -1,23 +1,26 @@
-Py_Type(x::PyPtr) = PyPtr(UnsafePtr(x).type[!])
+asptr(x) = Base.unsafe_convert(PyPtr, x)
 
-PyObject_Type(x::PyPtr) = (t = Py_Type(x); Py_IncRef(t); t)
+Py_Type(x) = Base.GC.@preserve x PyPtr(UnsafePtr(asptr(x)).type[!])
 
-Py_TypeCheck(o::PyPtr, t::PyPtr) = PyType_IsSubtype(Py_Type(o), t)
-Py_TypeCheckFast(o::PyPtr, f::Integer) = PyType_IsSubtypeFast(Py_Type(o), f)
+PyObject_Type(x) = Base.GC.@preserve x (t = Py_Type(asptr(x)); Py_IncRef(t); t)
 
-PyType_IsSubtypeFast(t::PyPtr, f::Integer) =
-    Cint(!iszero(UnsafePtr{PyTypeObject}(t).flags[] & f))
+Py_TypeCheck(o, t) = Base.GC.@preserve o t PyType_IsSubtype(Py_Type(asptr(o)), asptr(t))
+Py_TypeCheckFast(o, f::Integer) = Base.GC.@preserve o PyType_IsSubtypeFast(Py_Type(asptr(o)), f)
 
-PyMemoryView_GET_BUFFER(m::PyPtr) = Ptr{Py_buffer}(UnsafePtr{PyMemoryViewObject}(m).view)
+PyType_IsSubtypeFast(t, f::Integer) =
+    Base.GC.@preserve t Cint(!iszero(UnsafePtr{PyTypeObject}(asptr(t)).flags[] & f))
 
-PyType_CheckBuffer(t::PyPtr) = begin
-    p = UnsafePtr{PyTypeObject}(t).as_buffer[]
+PyMemoryView_GET_BUFFER(m) = Base.GC.@preserve m Ptr{Py_buffer}(UnsafePtr{PyMemoryViewObject}(asptr(m)).view)
+
+PyType_CheckBuffer(t) = Base.GC.@preserve t begin
+    p = UnsafePtr{PyTypeObject}(asptr(t)).as_buffer[]
     return p != C_NULL && p.get[!] != C_NULL
 end
 
-PyObject_CheckBuffer(o::PyPtr) = PyType_CheckBuffer(Py_Type(o))
+PyObject_CheckBuffer(o) = Base.GC.@preserve o PyType_CheckBuffer(Py_Type(asptr(o)))
 
-PyObject_GetBuffer(o::PyPtr, b, flags) = begin
+PyObject_GetBuffer(_o, b, flags) = Base.GC.@preserve _o begin
+    o = asptr(_o)
     p = UnsafePtr{PyTypeObject}(Py_Type(o)).as_buffer[]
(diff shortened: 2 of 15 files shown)


MilesCranmer:
@cjdoris you should try https://github.com/mxschmitt/action-tmate. It's pretty useful for when you don't have the OS that is failing

MilesCranmer:
I don't see the segfault btw. It just looks like a notebook error?

```python
_________________________ pytest/test_nb.ipynb::Cell 0 _________________________
Notebook cell execution failed
Cell 0: Timeout of 2000 seconds exceeded while executing cell. Failed to interrupt kernel in 5 seconds, so failing without traceback.

Input:
# NBVAL_IGNORE_OUTPUT
import numpy as np
```

cjdoris:
Oh goodie it's non-deterministic. The Mac test just passed but Linux failed! But Mac did fail with Python 3.13.2 (was previously testing on latest Python 3.13.3). At least I might be able to reproduce this locally on Linux...

cjdoris:
I have completely reverted this branch (except for some trivial details in the CI workflow definition) and still get the errors, so these should be fixed first. The error seems to only occur on Mac on Python 3.13.

MilesCranmer:
I have a mac but don't see the issue myself. Perhaps it was an upstream dependency which is now fixed?

MilesCranmer:
Ah sorry let me try Python 3.13, i was using 3.12.

MilesCranmer:
I can reproduce the segfault: I see it on 3.13.0-3.13.2, but _not_ on 3.13.3. So perhaps it was a Python bug that is now fixed?

cjdoris:
OK the current failures are due to https://github.com/JuliaLang/julia/issues/58171 so I guess we have another reproducer for that.

vchuravy:
> do you think the jl_object_id__cold might be coming from something fixable on the PythonCall.jl side, or is it definitely something in the Julia source itself?

I would say it's a high likelyhood that it is a missunderstanding between Python, PythonCall.jl and Julia's memory semantics. So not at all clear that it is in the Julia source itself.

MilesCranmer:
> > In fact I do wonder if that issue isn't a Julia bug (since we've never seen it anywhere besides the PythonCall.jl instance). Maybe it's due to how Python multithreading changed in 1.13 or something...
> 
> What changed?

I assumed there were some internal changes to enable the GIL free version: https://docs.python.org/3/library/threading.html. But actually I don't actually know if the regular GIL version had changes to enable this.

PR #625, opened 2025-06-10
docs: Code loading order

icweaver:
Added a note to help avoid the code loading edge case hit in https://github.com/JuliaPy/CondaPkg.jl/issues/188, e.g.,

```julia
using PythonCall
using CondaPkg
CondaPkg.add("numpy")
np = pyimport("numpy")
```

can fail on some platforms, while

```julia
using CondaPkg
CondaPkg.add("numpy")
using PythonCall
np = pyimport("numpy")
```

works as intended.


cjdoris:
cheers :)

PR #634, opened 2025-07-01
Add PySide6 support to gui.jl

### LINKED ISSUE #632: Support PySide6 in Compat
**Is your feature request related to a problem? Please describe.**
I am wrapping a library which depends on PySide6 and I always hit the error

```
qt.core.plugin.factoryloader: checking directory path "C:/Users/ejmei/.julia/juliaup/julia-1.11.2+0.x64.w64.mingw32/bin/platforms" ...
qt.qpa.plugin: Could not find the Qt platform plugin "windows" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
```

This output is with the `QT_DEBUG_PLUGINS` env variable turned on. Its looking for the platforms in definitely the incorrect place (the julia bin dir). I believe the function in Compat designed to fix this issue (or a similar issue), but does not support PySide6. It would also be nice to support the event loop, but I am less certain of what needs to be done to get that working.

There is a plugins directory here:
`.CondaPkg\.pixi\envs\default\Library\lib\qt6\plugins` 
and this is correctly set in the conf file, so I'm not really sure why qt is looking in my julia folder.

**Describe the solution you'd like**

The `fix_qt_plugin_path` function looks for the `qt.conf` file. For PySide6, this file seems to be called `qt6.conf` instead. I am not sure what the format of the other conf files are but I copied enough into the "addition context" section (including the relevant prefix variable). The existing function seems to just add `plugins` to the `Prefix` entry of the conf file. However that is not a folder is correctly specified in the conf (even though qt seems to be looking elsewhere). Maybe something else needs to change, I am not 100% sure.

**Additional context**

qt6.conf file contents
```
[Paths]                                                     
Prefix = D:/Code/repos/PyVITO/.CondaPkg/.pixi/envs/default                                      
Documentation = D:/Code/repos/PyVITO/.CondaPkg/.pixi/envs/default/Library/share/doc/qt6         
Headers = D:/Code/repos/PyVITO/.CondaPkg/.pixi/envs/default/Library/include/qt6                            
Libraries = D:/Code/repos/PyVITO/.CondaPkg/.pixi/envs/default/Library/lib
.... 
```

cjdoris:
This PR adds PySide6 support to the GUI compatibility layer.

## Changes Made

- **Updated `fix_qt_plugin_path() function** to support both Qt5 and Qt6 configurations:
  - Added support for qt6.conf files used by PySide6 (in addition to existing qt.conf support)
  - For Qt6, reads the Libraries entry instead of Prefix 
  - Sets QT_PLUGIN_PATH to {libraries}/qt6/plugins for PySide6 compatibility
  - Maintains backward compatibility with existing Qt5 frameworks

- **Added PySide6 to event loop callback** by including it in the supported GUI frameworks list and adding the import for PySide6.QtCore

- **Added PySide6 module hook** to automatically call `fix_qt_plugin_path() when PySide6 is imported

- **Updated documentation** to include PySide6 in all relevant docstrings and parameter descriptions

## Technical Details

The key insight from the issue discussion is that Qt6 uses a different configuration file format:
- Qt5: Uses qt.conf with Prefix entry, plugins at {prefix}/plugins
- Qt6: Uses qt6.conf with Libraries entry, plugins at {libraries}/qt6/plugins

The updated function checks for qt6.conf first, then falls back to qt.conf for backward compatibility.

## Testing

This change is designed to be backward-compatible and follows the exact approach suggested by @cjdoris in the issue comments. The implementation handles the platform-dependent path differences by using the Libraries entry from qt6.conf as recommended.

Fixes #632

### DIFF
--- a/src/Compat/gui.jl
+++ b/src/Compat/gui.jl
@@ -7,26 +7,49 @@ This fixes the problem that Qt does not know where to find its `qt.conf` file, b
 always looks relative to `sys.executable`, which can be the Julia executable not the Python
 one when using this package.
 
-If `CONFIG.auto_fix_qt_plugin_path` is true, then this is run automatically before `PyQt4`, `PyQt5`, `PySide` or `PySide2` are imported.
+If `CONFIG.auto_fix_qt_plugin_path` is true, then this is run automatically before `PyQt4`, `PyQt5`, `PySide`, `PySide2` or `PySide6` are imported.
 """
 function fix_qt_plugin_path()
     C.CTX.exe_path === nothing && return false
     e = pyosmodule.environ
     "QT_PLUGIN_PATH" in e && return false
-    qtconf = joinpath(dirname(C.CTX.exe_path::AbstractString), "qt.conf")
-    isfile(qtconf) || return false
-    for line in eachline(qtconf)
-        m = match(r"^\s*prefix\s*=(.*)$"i, line)
-        if m !== nothing
-            path = strip(m.captures[1]::AbstractString)
-            path[1] == path[end] == '"' && (path = path[2:end-1])
-            path = joinpath(path, "plugins")
-            if isdir(path)
-                e["QT_PLUGIN_PATH"] = realpath(path)
-                return true
+    
+    exe_dir = dirname(C.CTX.exe_path::AbstractString)
+    
+    # Check for Qt6 configuration first (PySide6)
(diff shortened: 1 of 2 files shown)


ejmeitz:
Seems to work for me. Was able to render this with PySide6

image

Issue #586, opened 2025-01-15
Segfault on propertynames


tecosaur:
**Affects:** PythonCall / JuliaCall / Both [delete as appropriate]

**Describe the bug**

Using Python 3.12.8 and Julia 1.11, when I type `<pyobject>.` into the REPL, I see a segfault.

**Your system**

```julia-repl
julia> versioninfo()
Julia Version 1.11.2
Commit 5e9a32e7af2 (2024-12-01 20:02 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 64 × AMD Ryzen Threadripper 3970X 32-Core Processor
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, znver2)
Threads: 60 default, 0 interactive, 30 GC (on 64 virtual cores)
Environment:
  JULIA_NUM_THREADS = 60
  JULIA_PROJECT = @.
  JULIA_EDITOR = emacs -nw

(...) pkg> st
Project HunFlair2 v0.1.0
Status `~/.../Project.toml`
  [992eb4ea] CondaPkg v0.2.24
  [6099a3de] PythonCall v0.9.23

(...) pkg> conda st
CondaPkg Status /.../CondaPkg.toml
Environment
  /.../.CondaPkg/env
Pip packages
  flair v0.14.0
```

**Additional context**

```julia-repl
julia> s.
[57673] signal 11 (1): Segmentation fault
in expression starting at none:0
_PyInterpreterState_GET at /usr/local/src/conda/python-3.12.8/Include/internal/pycore_pystate.h:133 [inlined]
_PyType_Lookup at /usr/local/src/conda/python-3.12.8/Objects/typeobject.c:4729 [inlined]
_PyObject_LookupSpecial at /usr/local/src/conda/python-3.12.8/Objects/typeobject.c:2167
_dir_object at /usr/local/src/conda/python-3.12.8/Objects/object.c:1758 [inlined]
PyObject_Dir at /usr/local/src/conda/python-3.12.8/Objects/object.c:1790
PyObject_Dir at /home/tec/.julia/packages/PythonCall/Nr75f/src/C/pointers.jl:303 [inlined]
macro expansion at /home/tec/.julia/packages/PythonCall/Nr75f/src/Core/Py.jl:132 [inlined]
[...]

PR #635, opened 2025-07-06
Race condition defences

MilesCranmer:
@cjdoris possible good news for https://github.com/JuliaLang/julia/issues/58171. See below.

Out of all of these changes, I _think_ the crucial one is `PYJLVALUES`. But the others are good defensive patterns so I think we should do it anyways.

I can now run 

```bash
python -c 'from juliacall import Main as jl; [jl.Ref(jl.randn(1)) for _ in range(1000000)]'
```

without getting a segfault on Python 3.13.

### DIFF
--- a/src/Convert/pyconvert.jl
+++ b/src/Convert/pyconvert.jl
@@ -12,8 +12,8 @@ struct PyConvertRule
     priority::PyConvertPriority
 end
 
-const PYCONVERT_RULES = Lockable(Dict{String,Vector{PyConvertRule}}())
-const PYCONVERT_EXTRATYPES = Lockable(Py[])
+const PYCONVERT_RULES = ErrorLockable(Dict{String,Vector{PyConvertRule}}())
+const PYCONVERT_EXTRATYPES = ErrorLockable(Py[])
 
 """
     pyconvert_add_rule(tname::String, T::Type, func::Function, priority::PyConvertPriority=PYCONVERT_PRIORITY_NORMAL)
@@ -261,7 +261,7 @@ function _pyconvert_get_rules(pytype::Py)
     return rules
 end
 
-const PYCONVERT_PREFERRED_TYPE = Lockable(Dict{Py,Type}())
+const PYCONVERT_PREFERRED_TYPE = ErrorLockable(Dict{Py,Type}())
 
 pyconvert_preferred_type(pytype::Py) =
     Base.@lock PYCONVERT_PREFERRED_TYPE get!(PYCONVERT_PREFERRED_TYPE[], pytype) do
@@ -307,7 +307,7 @@ end
 
 pyconvert_fix(::Type{T}, func) where {T} = x -> func(T, x)
 
-const PYCONVERT_RULES_CACHE = Lockable(IdDict{Any,Dict{C.PyPtr,Vector{Function}}}())
+const PYCONVERT_RULES_CACHE = ErrorLockable(IdDict{Any,Dict{C.PyPtr,Vector{Function}}}())
 
 function pyconvert_rules_cache(::Type{T}) where {T}
     Base.@lock PYCONVERT_RULES_CACHE get!(
--- a/src/JlWrap/C.jl
+++ b/src/JlWrap/C.jl
@@ -1,7 +1,7 @@
 module Cjl
 
 using ...C: C
-using ...Utils: Utils, Lockable
+using ...Utils: Utils, ErrorLockable
 using Base: @kwdef
 using UnsafePointers: UnsafePtr
 using Serialization: serialize, deserialize
@@ -16,7 +16,7 @@ const PyJuliaBase_Type = Ref(C.PyNULL)
 
 # we store the actual julia values here
 # the `value` field of `PyJuliaValueObject` indexes into here
-const PYJLVALUES = Lockable((; values=IdDict{Int,Any}(), free_slots=Int[], next_slot=Ref(1)))
+const PYJLVALUES = ErrorLockable((; values=IdDict{Int,Any}(), free_slots=Int[], next_slot=Ref(1)))
 
 function _pyjl_new(t::C.PyPtr, ::C.PyPtr, ::C.PyPtr)
     o = ccall(UnsafePtr{C.PyTypeObject}(t).alloc[!], C.PyPtr, (C.PyPtr, C.Py_ssize_t), t, 0)
@@ -39,7 +39,7 @@ function _pyjl_dealloc(o::C.PyPtr)
     nothing
 end
 
-const PYJLMETHODS = Lockable([])
+const PYJLMETHODS = ErrorLockable([])
 
 function PyJulia_MethodNum(f)
     @nospecialize f
@@ -65,7 +65,7 @@ function _pyjl_callmethod(o::C.PyPtr, args::C.PyPtr)
     return _pyjl_callmethod(f, o, args, nargs)::C.PyPtr
 end
 
-const PYJLBUFCACHE = Lockable(Dict{Ptr{Cvoid},Any}())
+const PYJLBUFCACHE = ErrorLockable(Dict{Ptr{Cvoid},Any}())
 
 @kwdef struct PyBufferInfo{N}
     # data
--- a/src/Convert/Convert.jl
+++ b/src/Convert/Convert.jl
@@ -9,7 +9,7 @@ using ..Core
 using ..Core:
     C,
     Utils,
-    Lockable,
+    ErrorLockable,
     @autopy,
     getptr,
     incref,
--- a/src/Core/Core.jl
+++ b/src/Core/Core.jl
@@ -11,7 +11,7 @@ const ROOT_DIR = dirname(dirname(@__DIR__))
 using ..PythonCall: PythonCall  # needed for docstring cross-refs
 using ..C: C
 using ..GC: GC
-using ..Utils: Utils, Lockable
+using ..Utils: Utils, ErrorLockable
 using Base: @propagate_inbounds, @kwdef
 using Dates:
     Date,
--- a/src/Core/Py.jl
+++ b/src/Core/Py.jl
@@ -56,7 +56,7 @@ decref(x::Py) = Base.GC.@preserve x (decref(getptr(x)); x)
 
 Base.unsafe_convert(::Type{C.PyPtr}, x::Py) = getptr(x)
 
-const PYNULL_CACHE = Lockable(Py[])
+const PYNULL_CACHE = ErrorLockable(Py[])
 
 """
     pynew([ptr])
--- a/src/Core/builtins.jl
+++ b/src/Core/builtins.jl
@@ -1206,7 +1206,7 @@ export pyfraction
 
 ### eval/exec
 
-const MODULE_GLOBALS = Lockable(Dict{Module,Py}())
+const MODULE_GLOBALS = ErrorLockable(Dict{Module,Py}())
 
 function _pyeval_args(code, globals, locals)
     if code isa AbstractString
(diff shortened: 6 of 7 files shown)


src/JlWrap/C.jl
```diff
@@ -339,22 +341,29 @@ end
...
-            idx = pop!(PYJLFREEVALUES)
-            PYJLVALUES[idx] = v
+        Base.@lock PYJLVALUES begin
+            if isempty(PYJLVALUES[].free_slots)
+                idx = PYJLVALUES[].next_slot[]
+                PYJLVALUES[].next_slot[] += 1
+            else
+                idx = pop!(PYJLVALUES[].free_slots)
+            end
+            PYJLVALUES[].values[idx] = v
```
MilesCranmer, line 360: This also seemed like a good thing to have... and might be relevant for why the segfault disappeared.

MilesCranmer:
@mkitti AFAICT the `GIL.@lock`/`GIL.@unlock` macros do not come into play in the segfault we were seeing so I am pretty sure these issues are orthogonal

mkitti:
You essentially have put a `ReentrantLock` on a bunch of data structures. Are the data structures themselves not thread safe?

These are Julian data structures, so this is unlikely.

What has been thread unsafe is Python and the GIL locking mechanisms.

Rather than creating a bunch of locks, we need one central one around the GIL locking mechanism.

MilesCranmer:
> You essentially have put a `ReentrantLock` on a bunch of data structures. Are the data structures themselves not thread safe?

`Array` is not thread safe for writes

> What has been thread unsafe is Python and the GIL locking mechanisms.
>
> Rather than creating a bunch of locks, we need one central one around the GIL locking mechanism.

Can you give an example? I am not sure I follow. There is no manual GIL manipulation in the segfault I am seeing. i.e., nowhere in the code path is there a `PyEval_SaveThread` as far as I can tell.

vtjnash (commented):
It looks like many of these functions would mandate holding the Python lock (GIL). Holding (or having) a julia lock also seems a recipe for deadlock and concurrency problems long term, though might fix some of the crashes short term

src/Convert/pyconvert.jl
```diff
@@ -307,10 +307,10 @@ end
 
 pyconvert_fix(::Type{T}, func) where {T} = x -> func(T, x)
 
-const PYCONVERT_RULES_CACHE = Dict{Type,Dict{C.PyPtr,Vector{Function}}}()
+const PYCONVERT_RULES_CACHE = Lockable(Dict{Type,Dict{C.PyPtr,Vector{Function}}}())
 
 @generated pyconvert_rules_cache(::Type{T}) where {T} =
```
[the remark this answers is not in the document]
MilesCranmer: Nice catch, thanks

src/Convert/pyconvert.jl
```diff
@@ -307,10 +307,11 @@ end
...
 pyconvert_fix(::Type{T}, func) where {T} = x -> func(T, x)
 
-const PYCONVERT_RULES_CACHE = Dict{Type,Dict{C.PyPtr,Vector{Function}}}()
+const PYCONVERT_RULES_CACHE = Lockable(Dict{Type,Dict{C.PyPtr,Vector{Function}}}())
 
-@generated pyconvert_rules_cache(::Type{T}) where {T} =
-    get!(Dict{C.PyPtr,Vector{Function}}, PYCONVERT_RULES_CACHE, T)
+pyconvert_rules_cache(::Type{T}) where {T} =
+    Base.@lock PYCONVERT_RULES_CACHE _pyconvert_rules_cache!(PYCONVERT_RULES_CACHE[], T)
+@generated _pyconvert_rules_cache!(d::Dict, ::Type{T}) where {T} = get!(Dict{C.PyPtr,Vector{Function}}, d, T)
```
MilesCranmer, line 314: I'm noticing this is probably illegal considering world ages (though maybe it's fine so long as nobody defines a new rule?). How important is the caching @cjdoris? Do we need it in a generated function?

mkitti:
Essentially, rather than adding a bunch of locks, let's just add one ReentrantLock in a central location so that only one Julia thread can communicate with Python at a time.

MilesCranmer:
Doesn't the Python GIL already result in this being true?

src/Convert/pyconvert.jl
vtjnash: It has always been illegal to access Dict from a generated function, since that implies mutation, which triggers many different forms of UB in inference all at once

MilesCranmer:
@dpinol did it work btw?

vtjnash:
This PR looks unsound, since you need to decide whether you want holding this lock to mean it is forbidden to access any Python state or if holding the Python lock means that it is forbidden to access any state protected by this lock. You cannot mix and match and hold both locks, unless you guarantee you always acquire the Python one first, in which case the Julia one is redundant and unnecessary. This is the "philosophers dining problem" of deadlock

MilesCranmer:
> holding this lock

Sorry, which lock do you mean?

visr:
The segfault from https://github.com/JuliaPy/PythonCall.jl/issues/586#issuecomment-3041416128 should be easy to verify if it was fixed. Currently this PR doesn't fix it. To try it, start Julia with a number of interactive threads != 1, and try to tab complete a Python object like `pyint`:

```jl
julia --threads=2,2

julia> using PythonCall

julia> x = pyint(3)
Python: 3

julia> x.<tab>
Please submit a bug report with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception: EXCEPTION_ACCESS_VIOLATION at 0x7ffadf016d2d -- PyType_Lookup at C:\ProgramData\DevDrives\.julia\dev\PythonCall\.CondaPkg\.pixi\envs\default\python312.dll (unknown line)
```

dpinol:
> @dpinol did it work btw?

After a couple of hours running with commit [AbstractLock undefined on 1.10](https://github.com/JuliaPy/PythonCall.jl/pull/635/commits/8d82e2a20b0854ce38d6e79ead13e56b905d23f7), no crash :partying_face: .
I'm trying now in parallel with latest commit 2eca3988ed58ab851cf9f170a4cec4c85513c7f0

MilesCranmer:
@dpinol great to hear!!

MilesCranmer:
If you're interested feel free to try. This PR is an imperfect but safe solution to a very specific segfault (`jl_object_id__cold`) when running Julia from Python via juliacall. This segfault has been impacting PySR and other Python libraries for over a year, basically preventing us from even _starting_ Julia on Python 3.13+. It is an extremely high priority fix. I'd be happy to help think about a more general approach down the road, but for now, others would need to push that direction, as I just don't have the time.

vtjnash:
Any function that accesses a Python object or Python state is not allowed to use a Julia lock as it must be using the GIL exclusively. To use a Julia lockable there may make those functions unsound (UB). This deserves some documentation in your PR to indicate which locks are permitted to be held at the same time and which are forbidden (UB) to acquire on the same thread at the same time

MilesCranmer:
@vtjnash I don't think these concerns apply here. This PR addresses a very specific and immediate segfault on the Python=>Julia side. These particular instances of lockables do not create a Julia-Python lock-ordering issue.

Currently, the data structures they guard get corrupted without the locking. This is a real, immediate, and severe problem, rather than something theoretical. To emphasise: Julia crashes _immediately_ on Python 3.13, on both macOS and Windows.

I fully support better long-term solutions (contributions welcome!), but I don't want to let perfect be the enemy of good here. I strongly prefer fixing the empirical segfault first, then tackling theoretical concerns afterward.

vtjnash:
In particular, the lock you added here to fix the crash appears to be the wrong lock to be using in the functions where it is used, since they seem to be direct calls into the python runtime (which require holding the GIL), so you are not actually fixing the cause of the segfault, but might be making it harder for someone to diagnose and fix correctly later (which appears to be in looking at the code to require holding the GIL, and which is what the package documentation says that all functions will do)
```
    ccall(UnsafePtr{C.PyTypeObject}(C.Py_Type(o)).free[!], Cvoid, (C.PyPtr,), o)
```

Since the documentation says every function should either already have or will acquire the GIL, it is strictly forbidden to introduce any new locks into this package (like this PR does), except for functions that do not interact with any python objects, since any lock acquired between the two recursive lock acquires would cause a violation of lock ordering.

MilesCranmer:
Can you illustrate your concerns with an example?

MilesCranmer:
I think worrying about API contracts when Julia won’t even start without a crash is like worrying about getting enough vegetables when your heart is stopped. Let’s do CPR first and worry about diet later.

cjdoris:
@MilesCranmer thanks for the PR, I'm glad we're closer to knowing what's going on.

I somewhat agree with @mkitti though, I don't think we should make every mutable global in sight into a `Lockable`. It makes the package slower and more complicated to maintain, and anyway I don't understand why this should be necessary. For now, can we just lock `PYJLVALUES` if that's the culprit for the current issue?

The fact that `Lockable` fixes things implies there is some race condition on `PYJLVALUES`, but I had thought that only one task or thread should be able to access this at a time anyway.

However, we now have an opportunity to diagnose the problem by inspecting activity of this lock. In particular, the times when we had to wait for the lock to be freed are the times there would have been illegal concurrent access. I suggest printing the stacktrace every time the lock is locked, and also whether or not we had to wait (which you can do with `trylock`).

I appreciate you want a quick fix but hopefully the above will shed some light on the underlying issue.

MilesCranmer:
I'm happy to try. However, regarding:

> _I don't think we should make every mutable global in sight into a Lockable_

I want to push back on this. I think defensive programming against thread races in this exact pattern is a very important software engineering habit. Julia data structures are not thread safe internally, so I believe it's incredibly important that any thread race should either (1) error loudly or be (2) safely avoided. The worst of the options, (3) is quiet corruption - which unlocked global mutables are always exposed to.

If code is already thread safe, then locking will have negligible performance effect - so there's no problem! But if code is not thread safe, then you really really _really_ want locks, otherwise we get the worst classes of bugs imaginable. (Which I have many scars to show for!) And even if code doesn't use multithreading on those parts now, someone might add it in the future. Or perhaps someone starts multithreading over nested PythonCall code in an outer loop somewhere. e.g., one time I segfaulted because I had two Wandb.jl loggers running in parallel in SymbolicRegression.jl (Wandb.jl uses PythonCall.jl internally) - who knows if they had simply hit one of these internal functions at the same time, or if it was something else.

If the performance is a worry, which I really do not think it would be be, I would recommend we make up some `ErrorLock` that simply errors whenever there is an attempted unlock from a second thread. That way we can be sure that any thread race, even if none are happening now, will be very obvious in the future.

mkitti:
What I think we might need to do is
1. Unlock the GIL on init by default as suggested by  @visr here:
https://github.com/JuliaPy/PythonCall.jl/issues/586#issuecomment-3042552597
2. Have every public or exported API function lock the GIL.
3. Provide non-locking versions of functions in a clearly marked module or perhaps even split this off into another package.

cjdoris:
@MilesCranmer OK yeah the `ErrorLock` looks good - just need to keep trying to reproduce the error to see what it reveals. I think it would be good to save off the stacktrace whenever it is locked too, and include it in the error, so that we can compare where locked the lock and where was trying to lock it.

I still can't reproduce on WSL (or anywhere locally) since that first day. 😞 Just need to keep trying...

dpinol:
I could not make it fail after multiple executions. I guess it depends on the machine load, because some days it crashes consistently, and others, it never fails.

MilesCranmer:
Interesting… so none of us can reproduce it today

dpinol:
> Interesting… so none of us can reproduce it today

I'll incorporate this check into my nightly CI, waiting for the day "the stars align" and it fails again.

MilesCranmer:
Going to re-open this in a fresh PR, now that we know the issue was with pointer casts.

Issue #627, opened 2025-06-19
A GC triggered while a thread is blocking on the GIL lock can freeze julia forever


NHDaly:
**Affects:** PythonCall

**Describe the bug**

The current implementation of allowing more than one Julia thread to directly attempt to lock the GIL, which can block the entire julia OS thread if the GIL is already locked, can hang julia forever because GC will fail to reach a safepoint.

Here is a very simple reproduction, adapted ever so slightly from the documentation:
- The docs suggest to run this example: https://juliapy.github.io/PythonCall.jl/stable/pythoncall/#jl-multi-threading
- Here, we adapt it by only adding a call to trigger GC while the GIL is held:
```julia
julia> using PythonCall

julia> PythonCall.GIL.@unlock Threads.@threads for i in 1:4
         PythonCall.GIL.@lock (GC.gc(); pyimport("time").sleep(5))
       end
..........................
.... hanging forever .....
...........................
```

**Your system**
```julia
julia> versioninfo()
Julia Version 1.10.2+RAI
Commit 52dcbad168* (2025-03-27 18:26 UTC)
Platform Info:
  OS: macOS (arm64-apple-darwin24.3.0)
  CPU: 12 × Apple M2 Max
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, apple-m1)
Threads: 4 default, 0 interactive, 4 GC (on 8 virtual cores)
Environment:
  JULIA_SSL_CA_ROOTS_PATH =

julia> Pkg.status()
Status `~/.julia/environments/v1.10/Project.toml`
  [69d22d85] About v1.0.2
  [1520ce14] AbstractTrees v0.4.5
  [6e4b80f9] BenchmarkTools v1.6.0
  [a2441757] Coverage v1.6.1
⌅ [f68482b8] Cthulhu v2.16.5
  [31a5f54b] Debugger v0.7.13
  [ab62b9b5] DeepDiffs v1.2.0
  [fb4d412d] FixedPointDecimals v0.6.3
  [c27321d9] Glob v1.3.1
  [92ed2492] HeapSnapshotUtils v0.1.0 `https://github.com/RelationalAI/HeapSnapshotUtils.jl#main`
  [7ec9b9c5] Humanize v1.0.0
  [5903a43b] Infiltrator v1.9.1
⌅ [70703baa] JuliaSyntax v0.4.10
[...]

PR #637, opened 2025-07-07
Surround the GIL with a ReentrantLock on the Julia side.

mkitti:
The Python GIL should only be able obtained by a single Julia thread / task.
To enforce this, we use a ReentrantLock on the Julia side that must be obtained
before trying to acquire the Python GIL.

Implements the fix suggested by https://github.com/JuliaPy/PythonCall.jl/issues/627 by @nhdaly and @kpamnany

### DIFF
--- a/src/GIL/GIL.jl
+++ b/src/GIL/GIL.jl
@@ -9,6 +9,18 @@ module GIL
 
 using ..C: C
 
+# Ensure that only one Julia thread tries to acquire the Python GIL
+# PyGILState_Ensure and PyGILState_Release may not be thread safe.
+# https://github.com/JuliaPy/PythonCall.jl/issues/627
+const _jl_gil_lock = ReentrantLock()
+
+"""
+    hasgil()
+
+Returns `true` if the current thread has the GIL or `false` otherwise.
+"""
+hasgil() = C.PyGILState_Check() == Cint(1)
+
 """
     lock(f)
 
@@ -21,11 +33,16 @@ threads. Since the main Julia thread holds the GIL by default, you will need to
 See [`@lock`](@ref) for the macro form.
 """
 function lock(f)
-    state = C.PyGILState_Ensure()
+    Base.lock(_jl_gil_lock)
     try
-        f()
+        state = C.PyGILState_Ensure()
+        try
+            f()
+        finally
+            C.PyGILState_Release(state)
+        end
     finally
-        C.PyGILState_Release(state)
+        Base.unlock(_jl_gil_lock)
     end
 end
 
@@ -42,11 +59,16 @@ The macro equivalent of [`lock`](@ref).
 """
 macro lock(expr)
     quote
-        state = C.PyGILState_Ensure()
+        Base.lock(_jl_gil_lock)
         try
-            $(esc(expr))
+            state = C.PyGILState_Ensure()
+            try
+                $(esc(expr))
+            finally
+                C.PyGILState_Release(state)
+            end
         finally
-            C.PyGILState_Release(state)
+            Base.unlock(_jl_gil_lock)
         end
     end
 end
@@ -63,11 +85,17 @@ Python code. That other thread can be a Julia thread, which must lock the GIL us
 See [`@unlock`](@ref) for the macro form.
 """
 function unlock(f)
-    state = C.PyEval_SaveThread()
+    _locked = Base.islocked(_jl_gil_lock)
+    _locked && Base.unlock(_jl_gil_lock)
     try
-        f()
+        state = C.PyEval_SaveThread()
+        try
+            f()
+        finally
+            C.PyEval_RestoreThread(state)
+        end
     finally
-        C.PyEval_RestoreThread(state)
+        _locked && Base.lock(_jl_gil_lock)
     end
 end
 
@@ -84,12 +112,25 @@ The macro equivalent of [`unlock`](@ref).
 """
 macro unlock(expr)
     quote
-        state = C.PyEval_SaveThread()
+        _locked = Base.islocked(_jl_gil_lock)
+        _locked && Base.unlock(_jl_gil_lock)
         try
-            $(esc(expr))
+            state = C.PyEval_SaveThread()
+            try
+                $(esc(expr))
+            finally
+                C.PyEval_RestoreThread(state)
+            end
         finally
-            C.PyEval_RestoreThread(state)
+            _locked && Base.lock(_jl_gil_lock)
         end
+    end
+end
+
+# If the main thread already has the GIL, we should lock _jl_gil_lock.
+function __init__()
+    if hasgil()
+        Base.lock(_jl_gil_lock)
     end
 end
 
--- a/test/GC.jl
+++ b/test/GC.jl
@@ -1,4 +1,7 @@
+using TestItemRunner
+
 @testitem "GC.gc()" begin
+    using PythonCall
     let
         pyobjs = map(pylist, 1:100)
         PythonCall.GIL.@unlock Threads.@threads for obj in pyobjs
@@ -13,6 +16,7 @@
 end
 
 @testitem "GC.GCHook" begin
+    using PythonCall
     let
         pyobjs = map(pylist, 1:100)
         PythonCall.GIL.@unlock Threads.@threads for obj in pyobjs
@@ -23,5 +27,10 @@ end
         VERSION >= v"1.10.0-" &&
         @test !isempty(PythonCall.GC.QUEUE.items)
     GC.gc()
+
+    # Unlock and relocking the ReentrantLock allows this test to pass
+    Base.unlock(PythonCall.GIL._jl_gil_lock)
+    Base.lock(PythonCall.GIL._jl_gil_lock)
+
     @test isempty(PythonCall.GC.QUEUE.items)
 end
--- a/test/GIL.jl
+++ b/test/GIL.jl
@@ -18,6 +18,14 @@
     if Threads.nthreads() ≥ 2
         @test 0.9 < t.time < 1.2
     end
+
+    @test PythonCall.GIL.hasgil()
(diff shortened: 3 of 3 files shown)


vtjnash:
I think this looks okay. The lock ordering will trigger nuisance errors when running with TSAN (since it usually local the Julia lock first except from init). It should be benign, but TSAN doesn't know that. You also probably should set the Task to sticky, since task thread migration while holding the GIL could also potentially trigger various sorts of memory errors. This PR seems like it should help avoid the potential issue that a task could mis-compute whether it actually has the GIL simply because a different task that ran on the same thread once had the GIL

src/GIL/GIL.jl
```diff
@@ -9,6 +9,18 @@ module GIL
 
 using ..C: C
 
+# Ensure that only one Julia thread tries to acquire the Python GIL
+# PyGILState_Ensure and PyGILState_Release may not be thread safe.
```
vtjnash, line 13: This reasoning is nonsense since those functions are explicitly thread safe. There may be good reasons for this PR, but this is not it.

vtjnash:
I think that test case is very wrong (currently). The only way I can think to make this sound and also allow that is to be more granular here (drop the locks whenever it isn't actively needed), but that seems borderline impossible with the current designs for Python

vtjnash:
That test should possibly be deadlocking if this PR implementation is correct. It is very difficult otherwise to show that you never violate the API contract here and attempt to acquire or hold the Julia lock at any time while still holding the GIL (because of the caller holding it recursively)

test/GC.jl
```diff
@@ -23,5 +27,10 @@ end
         VERSION >= v"1.10.0-" &&
         @test !isempty(PythonCall.GC.QUEUE.items)
     GC.gc()
+
+    # Unlock and relocking the ReentrantLock allows this test to pass
+    Base.unlock(PythonCall.GIL._jl_gil_lock)
+    Base.lock(PythonCall.GIL._jl_gil_lock)
```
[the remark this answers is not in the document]
vtjnash: finalizers must be disabled when any locks are held to avoid interrupting a critical section and causing memory corruption

mkitti:
My vague thought regarding https://github.com/JuliaPy/PythonCall.jl/pull/635 is that rather than creating individual `ReentrantLock`s for each data structure (`PYCONVERT_RULES`, `PYCONVERT_EXTRATYPES`, `PYNULL_CACHE`, `MODULE_GLOBALS`, and `PYJLVALUES`) they should at most just use one lock (`PythonCall.GIL._jl_gil_lock`).

src/GIL/GIL.jl
```diff
@@ -84,13 +145,29 @@ The macro equivalent of [`unlock`](@ref).
...
 end
 
+#=
+# Disable this for now since holding this lock will disable finalizers
+# If the main thread already has the GIL, we should lock _jl_gil_lock.
+function __init__()
+    if hasgil()
+        Base.lock(_jl_gil_lock)
+    end
+end
```
mkitti, line 170: Locking on `__init__` is probably a bad idea even though we are technically holding the GIL since this will disable finalizers.

Rather we should consider invoking `lock` while executing other calls.

mkitti: Disabling this apparently allowed the python tests to succeed.

MilesCranmer:
FYI this doesn't seem to fix https://github.com/JuliaLang/julia/issues/58171

```bash
bash> PYTHON_JULIACALL_HANDLE_SIGNALS=yes python -c 'from juliacall import Main as jl; [jl.Ref(jl.randn(1)) for _ in range(1000000)]'

[93492] signal 11 (2): Segmentation fault: 11
in expression starting at none:0
jl_object_id__cold at /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-honeycrisp-HL2F7YQ3XH.0/build/default-honeycrisp-HL2F7YQ3XH-0/julialang/julia-release-1-dot-11/src/builtins.c:441
```

mkitti:
> I guess it just indicates the GIL issue is independent of the juliacall segfaults?

No. You have not acquired the `_jl_gil_lock` yet.

What I'm suggesting is that you should acquire the same lock to protect the data structures. If there is one lock, the lock contention issue is greatly simplified. In theory, one independent task that acquires `_jl_gil_lock` should be able complete as long as it does not explicitly depend on the other tasks that are also trying to acquire the same lock.

mkitti:
> Ok I think I finally understand what you mean. So something like this?
> 
> ```diff
> + PYTHONCALL_STATE = Lockable((;
> - const PYJLVALUES = []
> +    values=[]
> - const PYJLFREEVALUES = Int[]
> +    free_values=Int[]
> - const PYJLMETHODS = Vector{Any}()
> +    methods=[]
> + ))
> ```
> 
> Rather than having separate constants throughout the library?

That's better since they all use the same lock. I think they probably should be the same lock as the GIL lock though.

```
const PYJLVALUES = Lockable([], GIL._jl_gil_lock)
const PYJLFREEVALUES = Lockable(Int[], GIL._jl_gil_lock)
const PYJLMETHODS = Lockable([], GIL._jl_gil_lock)
```

That is if a `Task` is holding the GIL then it should have free and exclusive access to all the structures.

`GIL._jl_gil_lock` is a `ReentrantLock` so the same `Task` can lock it multiple times as long as it unlocks it the same number of times.

```julia
julia> L = ReentrantLock();

julia> lock(L); lock(L); lock(L)

julia> @async begin
           lock(L)
           println("I got the lock")
       end
Task (runnable, started) @0x00000070aeedb650

julia> unlock(L)

julia> unlock(L)

julia> unlock(L)
I got the lock
```

Applied to this situation...

```
julia> const Lockable = Base.Lockable
Base.Lockable

julia> module GIL
           const _jl_gil_lock = ReentrantLock()
       end
Main.GIL

julia> const PYJLVALUES = Lockable([], GIL._jl_gil_lock);

julia> const PYJLMETHODS = Lockable([], GIL._jl_gil_lock);

julia> const PYJLFREEVALUES = Lockable(Int[], GIL._jl_gil_lock);

julia> lock(PYJLVALUES)

julia> lock(PYJLMETHODS)

julia> lock(PYJLFREEVALUES)

julia> @async begin
           @lock PYJLMETHODS println("I got the lock")
       end
Task (runnable, started) @0x0000007ed6063e10

julia> unlock(PYJLFREEVALUES)

julia> unlock(PYJLMETHODS)

julia> unlock(PYJLVALUES)
I got the lock
```

The problem is that we also want this same lock to represent holding the GIL. This is further complicated since a Julia `Task` can migrate across threads.

Per Jameson, it seems we may need to write up a custom lock that integrates the GIL locking mdchanism further.

MilesCranmer:
e.g., here: 

https://github.com/JuliaPy/PythonCall.jl/blob/c84c05954bf5e8da93e2e8efacd4177c032b0465/src/C/pointers.jl#L303

I think this ccall should be behind a lock, right?

src/GIL/GIL.jl
mkitti: `OncePerThread` is to be introduced in Julia 1.12 which has yet to be released at the time of this writing. Is there a lesser version of this that could be implemented prior to Julia 1.12, preferably to Julia 1.10 LTS?

MilesCranmer:
Yeah that makes sense to me. The reason I brought this up is my confusion over `PythonCall.GIL` here, because I thought this was a high-level interface for users to control the GIL, rather than something used internally by the package.

If you look at src/GC/GC.jl for example, it checks the GIL with a direct call `C.PyGILState_Check() == 1`, rather than using the GIL module. Perhaps src/GIL/GIL.jl needs to also hold a mechanism by which other parts of PythonCall can lock/unlock the GIL before executing C calls?

But then if this is true I don't understand why PythonCall.jl doesn't immediately crash when doing anything in a multithreaded context. So clearly my mental model for this is wrong somehow.

@cjdoris how do the C calls ensure they don't execute Python bytecode in parallel? Is the GIL itself preventing it or something?

src/GIL/GlobalInterpreterLock.jl
```diff
@@ -0,0 +1,120 @@
...
+        if checkbounds(Bool, xs, tid)
+            if ss[tid] == 0
+                xs[tid] = once.initializer()
+                ss[tid] = 1
+            end
+            return xs[tid]
+        else
+            throw(ErrorException("Thread id $tid is out of bounds as initially allocated. Use Julia 1.12 for dynamic thread usage."))
+        end
+    end
```
MilesCranmer, line 92: Why not make this per-task rather than per-thread?

I use something like this in SymbolicRegression.jl:

```julia
struct PerTaskCache{T,F<:Function}
    constructor::F
    PerTaskCache{T}(constructor::F) where {T,F} = new{T,F}(constructor)
    PerTaskCache{T}() where {T} = PerTaskCache{T}(() -> T())
end
function Base.getindex(cache::PerTaskCache{T}) where {T}
    tls = Base.task_local_storage()
    haskey(tls, cache) && return tls[cache]::T
    value = cache.constructor()::T
    tls[cache] = value
    return value
end
function Base.setindex!(cache::PerTaskCache{T}, value::T) where {T}
    tls = Base.task_local_storage()
    tls[cache] = value
    return value
end
```

Then I can use it on each task:

```julia
const CachedPrep = PerTaskCache{Dict{UInt,Any}}()

function foo()
    cache = CachedPrep[]
    #= do stuff with cache::Dict{UInt,Any} =#
end
```

mkitti:
I've been testing this on Julia 1.11 and Julia 1.12. On Julia 1.12 this is extremely unstable. It seems to work fine on Julia 1.11 though.

src/GIL/GIL.jl
vtjnash: Yes, Base.Condition, since this is locked to thread-local state

PR #640, opened 2025-07-15
hotfix: misaligned pointer cast

MilesCranmer:
This isn't the full solution from the discourse thread, but I would like to get a hotfix in ASAP to prevent further segfaults until we can get a future-compatible heap type solution in.

@cjdoris could you merge and release?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "PythonCall"
 uuid = "6099a3de-0909-46bc-b1f4-468b9a2dfc0d"
 authors = ["Christopher Doris <github.com/cjdoris>"]
-version = "0.9.25"
+version = "0.9.26"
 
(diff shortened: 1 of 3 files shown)


dpinol:
Would this work for Python 3.11 & 3.12 if it specifies fields only on these versions?

MilesCranmer:
I think so. The worse issue is having fewer fields, because then it will overwrite neighboring constants and corrupt them. Correct me if I'm wrong @cjdoris

MilesCranmer:
I think benchmark pr failing is unrelated (and likely due to a slightly older version of my fork). Looks ready for merge though @cjdoris (thanks for the quick help)

cjdoris:
Released. Hopefully that's the segfault gone!

PR #644, opened 2025-07-17
Switch to heap types

### LINKED ISSUE #641: Switch to heap types
This is an issue to track implementation of the following helpful suggestion by @zerointensity on [discourse](https://discourse.julialang.org/t/plea-for-segfault-assistance-in-pythoncall-jl/130697/37?u=milescranmer)

> Alternatively, it’d be nice if Julia didn’t have to use the PyTypeObject structure at all. I’m assuming you guys are using it to define native types? If so, I think it’d be a lot easier to switch to heap types (PyType_Spec), which use an array of slots instead of structure fields to avoid ABI compatibility issues.

cc @cjdoris 

Hotfix is now master, to be released soon, though it’s just a temporary fix to stop the bleeding until a full switch can be implemented.

MilesCranmer:
@cjdoris 

Fixes #641


cjdoris:
Closing as duplicate of #643 (sorry!) but note https://github.com/JuliaPy/PythonCall.jl/issues/641#issuecomment-3084603157

PR #649, opened 2025-07-22
[RFC] Module scripts

klamike:
This PR adds two module scripts:
- `python -m juliacall` / `python -m juliacall.repl` to launch a REPL (with a custom banner 😄)
- `python -m juliacall.init` to (lazily) download Julia and optionally show juliapkg env info.

The idea is these are nice for debugging -- calling `python -m juliacall` with the same exact environment as `python juliacall_script.py`, you know you are running exactly the same julia and project. The `juliacall.init` is nice as a quick sanity check, and in distributed settings like [discourse post](https://discourse.julialang.org/t/precompilation-gridlock-on-hpc-cluster/114213).

Happy to make any changes!

Here's what the REPL looks like:
Screenshot 2025-07-22 at 7 36 02 PM

### DIFF
--- a/pysrc/juliacall/repl.py
+++ b/pysrc/juliacall/repl.py
@@ -1,40 +1,31 @@
-def run_repl():
+def run_repl(banner='yes', quiet=False, history_file='yes', preamble=None):
     import os
     os.environ.setdefault("PYTHON_JULIACALL_HANDLE_SIGNALS", "yes")
     if os.environ.get("PYTHON_JULIACALL_HANDLE_SIGNALS") != "yes":
         print("Experimental JuliaCall REPL requires PYTHON_JULIACALL_HANDLE_SIGNALS=yes")
         exit(1)
 
-    import argparse
-    from pathlib import Path
-    parser = argparse.ArgumentParser("JuliaCall REPL (experimental)")
-    parser.add_argument('--banner', choices=['yes', 'no', 'short'], default='yes', help='Enable or disable startup banner')
-    parser.add_argument('--quiet', '-q', action='store_true', help='Quiet startup: no banner, suppress REPL warnings')
-    parser.add_argument('--history-file', choices=['yes', 'no'], default='yes', help='Load or save history')
-    parser.add_argument('--preamble', type=Path, help='Code to be included before the REPL starts')
-    args = parser.parse_args()
-
     from juliacall import Main, Base
 
     Base.is_interactive = True
 
-    if not args.quiet:
+    if not quiet:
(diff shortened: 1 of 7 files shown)


pysrc/juliacall/repl.py
```diff
@@ -0,0 +1,45 @@
...
+        print("Experimental JuliaCall REPL requires PYTHON_JULIACALL_HANDLE_SIGNALS=yes")
+        exit(1)
+
+    Base.is_interactive = True
+
+    if not quiet:
+        Main.include(os.path.join(os.path.dirname(__file__), 'banner.jl'))
+        Main.__PythonCall_banner(Base.Symbol(banner))
+
+    if Main.seval(r'VERSION > v"v1.11.0-alpha1"'):
```
klamike, line 19:
```suggestion
    if Main.seval(r'VERSION ≥ v"v1.11.0-alpha1"'):
```

pytest/test_all.py
```diff
@@ -168,3 +168,45 @@ def test_call_nogil(yld, raw):
...
+    import juliacall as _
+    import subprocess
+    import time
+
+    jl_version = juliapkg.state.STATE["version"]
+
+    # grab PythonCall.jl version from pyproject.toml
+    with open("pyproject.toml", "rb") as f:
+        pyproject = tomllib.load(f)
+    pythoncall_version = pyproject["project"]["version"]
```
klamike, line 186: Not sure if this is always correct -- is the `juliacall` version always equal to the `PythonCall.jl` version?

cjdoris:
Hello, thanks for the PR. If I'm honest I don't think JuliaCall would benefit from a CLI.

Can you explain why being able to launch a Julia REPL like this is useful? I think a better alternative would be to add a CLI to JuliaPkg to run Julia directly, like `python -m juliapkg run-julia args...`.

I do see the utility of the init CLI but again I think it would be better in JuliaPkg like `python -m juliapkg resolve`.

In conclusion, I'd totally be up for a PR to add a CLI to JuliaPkg with the above commands (and others like `add`, `rm` and `status`).

cjdoris:
On the discoverability side, juliapkg is in the docs for juliacall, which should suffice. I'm always happy to take PRs to improve the docs.

I'm happy to have the REPL stuff but over in juliacall like I said, which would directly launch Julia as a subprocess, instead of starting a REPL from within Julia.

If you do make a PR on juliapkg, please use click to create the CLI, using click's subcommands instead of separate modules.

I'll close this PR. Feel free to open an issue on juliapkg if you want to discuss further.

PR #666, opened 2025-09-13
NumpyDates

cjdoris:
- Adds a public NumpyDates submodule with types:
  - AbstractDateTime64, DateTime64, InlineDateTime64
  - AbstractTimeDelta64, TimeDelta64, InlineTimeDelta64
- Adds conversions from `numpy.datetime64` and `numpy.timedelta64` to these types.
- Adds conversions from arrays of these to `PyArray`.
- Tests for all of the above.

Todo:
- [x] Tests for NumpyDates.
- [ ] Tests for conversion of numpy scalars to these types.
- [ ] Tests for conversion of numpy arrays to these types.
- [x] Update changelog.
- [x] Update docs.


Beforerr:
Happy to see the support for NumpyDates. 

I would like to share a conversion function for  `datetime64[ns]` I previously used. It is much faster than `PyArray`. So maybe it is worth to put into the package.

```
function pyconvert_time(times)
    len = length(times)
    len == 0 && return UnixTime[]
    py_ns = PyArray{Int64}(@py times.view("i8"); copy = false)
    return reinterpret(UnixTime, py_ns)
end
```

```
julia> @b PyArray(var.py.time; copy=false)
941.625 μs (56 allocs: 1.828 KiB)
```

```
julia> @b Speasy.pyconvert_time(var.py.time)
670.710 ns (9 allocs: 416 bytes)
```

PR #667, opened 2025-09-13
Add options for the Julia binary and project path

JamesWrigley:
This makes using JuliaPkg optional. Also sneaked in a change to use `[sources]` in the docs project.

Tested manually. Should fix https://github.com/JuliaPy/pyjuliapkg/issues/56.


docs/Project.toml
```diff
@@ -4,3 +4,6 @@ PythonCall = "6099a3de-0909-46bc-b1f4-468b9a2dfc0d"
 
 [compat]
 Documenter = "1"
+
+[sources]
+PythonCall = {path = ".."}
```
cjdoris, line 9: 👍

docs/src/juliacall.md
```diff
@@ -125,6 +135,8 @@ be configured in two ways:
 | `-X juliacall-warn-overwrite=<yes\|no>` | `PYTHON_JULIACALL_WARN_OVERWRITE=<yes\|no>` | Enable or disable method overwrite warnings. |
 | `-X juliacall-autoload-ipython-extension=<yes\|no>` | `PYTHON_JULIACALL_AUTOLOAD_IPYTHON_EXTENSION=<yes\|no>` | Enable or disable IPython extension autoloading. |
 | `-X juliacall-heap-size-hint=<N>` | `PYTHON_JULIACALL_HEAP_SIZE_HINT=<N>` | Hint for initial heap size in bytes. |
+| `-X juliacall-exepath=<file>` | `PYTHON_JULIACALL_EXEPATH=<file>` | Path to Julia binary to use (overrides JuliaPkg). |
```
cjdoris, line 138: Should be `exe` not `exepath`.

pysrc/juliacall/__init__.py
```diff
@@ -147,21 +147,32 @@ def args_from_config(config):
...
-    # Find the Julia executable and project
-    CONFIG['exepath'] = exepath = juliapkg.executable()
-    CONFIG['project'] = project = juliapkg.project()
+        # Find the Julia executable and project
+        CONFIG['exe'] = juliapkg.executable()
+        CONFIG['project'] = juliapkg.project()
+    elif (not have_exe and have_project) or (have_exe and not have_project):
+        raise Exception("Both PYTHON_JULIACALL_PROJECT and PYTHON_JULIACALL_EXE must be set together, not only one of them.")
+
+    exe = CONFIG['exe']
```
cjdoris, line 171: Please revert to using `exepath` internally, like `exepath = CONFIG['exepath']`.

.github/workflows/tests.yml
```diff
@@ -82,10 +92,18 @@ jobs:
           python-version: ${{ matrix.pyversion }}
   
       - name: Set up Julia
+        id: setup_julia
+        if: ${{ matrix.juliaexe == 'julia' }}
```
cjdoris, line 96: Please do this unconditionally as before.

cjdoris (changes requested):
Looking good just a couple of small tweaks please.

docs/src/juliacall.md
JamesWrigley: Oops, fixed in ac95036.

.github/workflows/tests.yml
JamesWrigley: Fixed in ac95036. But why do we want to install it unconditionally?

cjdoris:
I made a few little tweaks before merging. Great work, thank you!

PR #698, opened 2025-10-18
Delete unnecessary comparison operators

JamesWrigley:
These are already defined by Base in terms of other comparison operators, so they cause a ton of invalidations. I believe the Python operators have the same logical requirements as the corresponding Julia ones, so this should still be correct :crossed_fingers: 

The total number of invalidations on Julia 1.12 went down to 8551 from 11647 (not including the fixed invalidations from the UnsafePointers.jl and Julia PRs, those remove a few more thousand). These invalidations should be fixed:
```julia
 inserting >=(x::PythonCall.Py, y::PythonCall.Py) @ PythonCall.Core ~/git/PythonCall.jl/src/Core/Py.jl:370 invalidated:                                                                                                                        
   backedges: 1: superseding >=(x, y) @ Base operators.jl:472 with MethodInstance for >=(::Any, ::Any) (2 children)

 inserting >=(x::Number, y::PythonCall.Py) @ PythonCall.Core ~/git/PythonCall.jl/src/Core/Py.jl:389 invalidated:                                                                                                                               
   backedges: 1: superseding >=(x, y) @ Base operators.jl:472 with MethodInstance for >=(::Int64, ::Any) (10 children)

 inserting >(x::Number, y::PythonCall.Py) @ PythonCall.Core ~/git/PythonCall.jl/src/Core/Py.jl:390 invalidated:
   backedges: 1: superseding >(x, y) @ Base operators.jl:425 with MethodInstance for >(::Int64, ::Any) (2 children)
              2: superseding >(x, y) @ Base operators.jl:425 with MethodInstance for >(::Int64, ::Any) (43 children)

 inserting !=(x::Number, y::PythonCall.Py) @ PythonCall.Core ~/git/PythonCall.jl/src/Core/Py.jl:386 invalidated:
   backedges: 1: superseding !=(x, y) @ Base operators.jl:321 with MethodInstance for !=(::Int64, ::Any) (69 children)

 inserting >(x::PythonCall.Py, y::PythonCall.Py) @ PythonCall.Core ~/git/PythonCall.jl/src/Core/Py.jl:371 invalidated:
   backedges: 1: superseding >(x, y) @ Base operators.jl:425 with MethodInstance for >(::Any, ::Any) (225 children)

 inserting !=(x::PythonCall.Py, y::PythonCall.Py) @ PythonCall.Core ~/git/PythonCall.jl/src/Core/Py.jl:367 invalidated:
   backedges: 1: superseding !=(x, y) @ Base operators.jl:321 with MethodInstance for !=(::Any, ::Any) (377 children)

 inserting >(x::PythonCall.Py, y::Number) @ PythonCall.Core ~/git/PythonCall.jl/src/Core/Py.jl:381 invalidated:
   backedges: 1: superseding >(x, y) @ Base operators.jl:425 with MethodInstance for >(::Any, ::Bool) (8 children)
              2: superseding >(x, y) @ Base operators.jl:425 with MethodInstance for >(::Any, ::Int128) (8 children)
              3: superseding >(x, y) @ Base operators.jl:425 with MethodInstance for >(::Any, ::UInt128) (8 children)
              4: superseding >(x, y) @ Base operators.jl:425 with MethodInstance for >(::Any, ::Int64) (8 children)
              5: superseding >(x, y) @ Base operators.jl:425 with MethodInstance for >(::Any, ::UInt64) (17 children)
              6: superseding >(x, y) @ Base operators.jl:425 with MethodInstance for >(::Any, ::UInt8) (56 children)
              7: superseding >(x, y) @ Base operators.jl:425 with MethodInstance for >(::Any, ::Integer) (66 children)
              8: superseding >(x, y) @ Base operators.jl:425 with MethodInstance for >(::Any, ::Int64) (209 children)

 inserting >=(x::PythonCall.Py, y::Number) @ PythonCall.Core ~/git/PythonCall.jl/src/Core/Py.jl:380 invalidated:
   backedges: 1: superseding >=(x, y) @ Base operators.jl:472 with MethodInstance for >=(::Any, ::Int64) (2 children)
              2: superseding >=(x, y) @ Base operators.jl:472 with MethodInstance for >=(::Any, ::UInt64) (3 children)
              3: superseding >=(x, y) @ Base operators.jl:472 with MethodInstance for >=(::Any, ::Int64) (408 children)

 inserting !=(x::PythonCall.Py, y::Number) @ PythonCall.Core ~/git/PythonCall.jl/src/Core/Py.jl:377 invalidated:
   backedges: 1: superseding !=(x, y) @ Base operators.jl:321 with MethodInstance for !=(::Any, ::Bool) (14 children)
              2: superseding !=(x, y) @ Base operators.jl:321 with MethodInstance for !=(::Any, ::UInt8) (32 children)
              3: superseding !=(x, y) @ Base operators.jl:321 with MethodInstance for !=(::Any, ::Int64) (375 children)
```

### DIFF
--- a/test/GIL.jl
+++ b/test/GIL.jl
@@ -3,7 +3,7 @@
     # GIL, these can happen in parallel if Julia has at least 2 threads.
     function threaded_sleep()
         PythonCall.GIL.unlock() do
-            Threads.@threads for i = 1:2
+            Threads.@threads :static for i = 1:2
                 PythonCall.GIL.lock() do
                     pyimport("time").sleep(1)
                 end
@@ -24,7 +24,7 @@ end
     # This calls Python's time.sleep(1) twice concurrently. Since sleep() unlocks the
     # GIL, these can happen in parallel if Julia has at least 2 threads.
     function threaded_sleep()
-        PythonCall.GIL.@unlock Threads.@threads for i = 1:2
+        PythonCall.GIL.@unlock Threads.@threads :static for i = 1:2
             PythonCall.GIL.@lock pyimport("time").sleep(1)
         end
     end


cjdoris:
This is a tricky one because on the one hand, these definitions do need to stay as they are because Python is weird. For example `x == y` in Python doesn't have to return a `bool`, and certainly `(x != y)` is allowed to be different from `not (x == y)`.

On the other hand, these comparison operators return a `Py` which violates the docstring for `==` (which I only just read and which explicitly says the result is a `Bool` unless one argument is a `Missing`) so these are illegal methods to start with. They are puns, really.

One can view this latter point as a bug and therefore fixable in a technically non-breaking fashion. But it definitely will break people's code. We're working (extremely slowly) towards PythonCall v1 which can definitely fix these issues.

The right thing to do would be to replace all these methods with ones that correctly return `Bool` (not `Py`) and remove the unnecessary ones in this PR (hence using the generic fallbacks in Base). I'll ponder if we can/should do this pre-v1.

cjdoris:
Thinking some more, actually I think these methods are not documented except there is one example which implies that `==(::Py, ::Int)` returns `Py`.

So perhaps it's ok to change these methods in a non-breaking release since the behaviour is (almost) undocumented, and is incompatible with the definition of `Base.==` anyway.

However like I said, this change will quite likely break a bunch of code in practice.

PR #716, opened 2025-11-11
Implement Base.IndexStyle for PyObjectArray using a type argument

JamesWrigley:
Base already defines a IndexStyle(::AbstractArray) method for object arguments that uses the object type, so we can implement that directly and avoid the invalidations from implement Base.IndexStyle(::PyObjectArray). Fixes these invalidations:
```julia
 inserting IndexStyle(x::PythonCall.PyObjectArray) @ PythonCall.JlWrap ~/.julia/dev/PythonCall/src/JlWrap/objectarray.jl:13 invalidated:                                                                                                       
   backedges: 1: superseding IndexStyle(A::AbstractArray) @ Base indices.jl:94 with MethodInstance for IndexStyle(::AbstractArray) (132 children) 
```

### DIFF
--- a/src/JlWrap/objectarray.jl
+++ b/src/JlWrap/objectarray.jl
@@ -10,7 +10,7 @@ PyObjectArray(x::AbstractArray{T,N}) where {T,N} = PyObjectArray{N}(x)
 
 pyobjectarray_finalizer(x::PyObjectArray) = GC.enqueue_all(x.ptrs)
 
-Base.IndexStyle(x::PyObjectArray) = Base.IndexStyle(x.ptrs)
+Base.IndexStyle(T::Type{<: PyObjectArray}) = Base.IndexStyle(fieldtype(T, :ptrs))
 
 Base.length(x::PyObjectArray) = length(x.ptrs)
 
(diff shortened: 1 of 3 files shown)


JamesWrigley:
I took the liberty of removing some `Dates.value()` methods in 7766cc8 that seem unnecessary, on the Julia 1.12.2 backports branch that fixes these invalidations:
```julia
 inserting value(d::PythonCall.NumpyDates.TimeDelta64) @ PythonCall.NumpyDates /gpfs/exfel/data/scratch/wrigleyj/julia-depot/packages/PythonCall/MxsO4/src/NumpyDates/TimeDelta64.jl:26 invalidated:
   backedges: 1: superseding value(x::Dates.Period) @ Dates ~/src/julia/usr/share/julia/stdlib/v1.12/Dates/src/periods.jl:10 with MethodInstance for Dates.value(::Dates.Period) (285 children)
```

Bringing the total number of invalidations for v1 on the future 1.12.2 release down to 340 :)

cjdoris:
Oh the tests aren't running because changing the base doesn't trigger them... Can you push a trivial commit (or two commits to get back to the current state) to trigger it? It'll all get squashed away anyway. Or maybe there is a more elegant way.

JamesWrigley:
I squashed them in e3e4f6a, that seems to have done the trick.

PR #718, opened 2025-11-12
Fix arguments for `ArrayValue.__array__()`

JamesWrigley:
Not sure how but this was broken on the v1 branch. Maybe from a rebase mistake?

### DIFF
--- a/src/JlWrap/array.jl
+++ b/src/JlWrap/array.jl
@@ -357,8 +357,14 @@ function init_array()
                     # convert to an array-like object
                     arr = self
                     if not (hasattr(arr, "__array_interface__") or hasattr(arr, "__array_struct__")):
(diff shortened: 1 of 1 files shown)


JamesWrigley:
I found some other issues with `__array__()`:
- The `copy=None` argument is not supported on numpy v1, tried fixing that with a poor mans version check in 9ac67bd.
- It always returns a numpy array with `dtype=object`. I suspect this is because of the missing 'first attempt' in v1 compared to main, which I tried adding back in 31e9a8c.

Not entirely sure if my changes are correct though.

cjdoris:
Can you make a separate issue about numpy v1 please.

JamesWrigley:
Sorry, forgot about this :sweat_smile:

PR #723, opened 2025-11-14
remove support for python 3.14

cjdoris:
see the updated faq for why

### DIFF
--- a/CondaPkg.toml
+++ b/CondaPkg.toml
@@ -10,7 +10,7 @@ version = "<=julia"
 
 [deps.python]
 build = "**cpython**"
-version = ">=3.10,<4"
+version = ">=3.10,<3.14"
 
 [dev.deps]
 matplotlib = ""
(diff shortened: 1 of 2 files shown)


PallHaraldsson:
Your version restriction in PythonCall.jl applies to juliacall automatically I believe.

Are you sure the problem also applied to calling from Python? I asked the Google search AI, and it it seemed to agree (though I wasn't sure either way). I think the support restriction should also go into the README.md; because it only applies to last version, right? It's a bit hard (impossible?) to do retroactively, and the older versions will error. What the AI told me (presumably Gemini, maybe latest 3 (which claims and seems great), though likely still using older Gemini 2.5 flash?):

>JuliaCall (the Python library counterpart) is also broken by this change for the same fundamental reasons. It's an issue with the underlying mechanism of calling back and forth between the two runtimes when using stack switching. The issue is a core one that the Python core developers are discussing, and a fix is not trivial as it involves assumptions about the C ABI and stack pointers.

PR #755, opened 2026-03-20
Implement a pyimport string macro

JamesWrigley:
This helps usability by automatically converting Python import statements into their Julia equivalents.

Example from the docs:
```julia
pyimport"""
import numpy as np
from scipy import linalg, optimize
from os.path import join as pathjoin
"""

# Converted to:
const np = pyimport("numpy")
const linalg = pyimport("scipy" => "lingalg")
const optimize = pyimport("scipy" => "optimize")
const pathjoin = pyimport("os.path" => "join")
```

### DIFF
--- a/docs/src/pythoncall-reference.md
+++ b/docs/src/pythoncall-reference.md
@@ -55,6 +55,7 @@ pyhasitem
 pyhash
 pyhelp
 pyimport
+@pyimport_str
 pyin
 pyis
 pyisinstance
--- a/src/API/exports.jl
+++ b/src/API/exports.jl
@@ -50,6 +50,7 @@ export pyilshift
 export pyimatmul
 export pyimod
 export pyimport
+export @pyimport_str
 export pyimul
 export pyin
 export pyindex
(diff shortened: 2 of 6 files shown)


JamesWrigley:
(bump)

JamesWrigley:
(bump)

JamesWrigley:
(bump)

cjdoris:
Hi sorry just getting back to my inbox after a long while!

What's this useful for? What's wrong with just using `pyimport(...)`?

Also I believe you can achieve the same thing with `@py` except with Julia syntax, like

```julia
@py begin
    import numpy as np
    import scipy: linalg, optimize
    import os.path: join as pathjoin
end
```

JamesWrigley:
No worries :)

It's purely to make importing Python libraries easier, I always found that translating import statements from Python syntax into what `pyimport()` expects to be a bit annoying. But `@py` looks super cool, I didn't realize that also handled import statements :eyes: The only thing that I'm missing is `from`, do you think it'd be possible to support that too?

JamesWrigley:
NVM, I see that it has to go through Julia's parser through so it's probably not possible. But I think `@py` is fine anyway for my use case so I'll close this.

Issue #669, opened 2025-09-16
rare ijl_bounds_error_int  with Julia 1.11 & python 3.13


dpinol:
**Affects:** JuliaCall

**Describe the bug**
Very rarely, my python application crashes while  juliacall is invoking a julia function.

This is the error I get:
```
error in running finalizer: BoundsError(a=MemoryRef(offset=0, ptr_or_offset=0x5a44f85dd300, mem=Memory{Int64}(1819, 0x5a44f85dd300)[....]

ijl_bounds_error_int at /cache/build/tester-amdci4-12/julialang/julia-release-1-dot-11/src/rtutils.c:207
memoryref at ./boot.jl:524 [inlined]
unsafe_copyto! at ./genericmemory.jl:115 [inlined]
#133 at ./array.jl:1133
_growend! at ./array.jl:1116 [inlined]
push! at ./array.jl:1261 [inlined]
_pyjl_dealloc at /home/avatar/.julia/packages/PythonCall/hXycg/src/JlWrap/C.jl:35
jfptr__pyjl_dealloc_9930 at /home/avatar/.julia/compiled/v1.11/PythonCall/WdXsa_ttmfG.so (unknown line)
jlcapi__pyjl_dealloc_10065 at /home/avatar/.julia/compiled/v1.11/PythonCall/WdXsa_ttmfG.so (unknown line)
subtype_dealloc at /tmp/python-build.20250826113541.3096764/Python-3.13.7/Objects/typeobject.c:2388
Py_DecRef at /home/avatar/.julia/packages/PythonCall/hXycg/src/C/pointers.jl:300 [inlined]
enqueue at /home/avatar/.julia/packages/PythonCall/hXycg/src/GC/GC.jl:99
py_finalizer at /home/avatar/.julia/packages/PythonCall/hXycg/src/Core/Py.jl:27
unknown function (ip: 0x7c5556d13607)
run_finalizer at /cache/build/tester-amdci4-12/julialang/julia-release-1-dot-11/src/gc.c:303
jl_gc_run_finalizers_in_list at /cache/build/tester-amdci4-12/julialang/julia-release-1-dot-11/src/gc.c:393
run_finalizers at /cache/build/tester-amdci4-12/julialang/julia-release-1-dot-11/src/gc.c:439
run_finalizers at /cache/build/tester-amdci4-12/julialang/julia-release-1-dot-11/src/gc.c:420 [inlined]
[...]

PR #758, opened 2026-04-16
fix crashes using multithreaded julia from python

dpinol:
Solves https://github.com/JuliaPy/PythonCall.jl/issues/669

The vectors PYJLVALUES and PYJLFREEVALUES are modified without any protection against GC-triggered re-entrancy.

```
  _pyjl_dealloc(o1) or PyJuliaValue_SetValue
    └→ push!(PYJLFREEVALUES, idx)          # begins resize (sets internal flag)
         └→ vector must grow → allocates → triggers Julia GC
              └→ GC collects a Py object → runs py_finalizer
                   └→ enqueue(ptr) → PyGILState_Check()==1 (same thread holds GIL)
                        └→ Py_DecRef(ptr) → refcount hits 0
                             └→ _pyjl_dealloc(o2)
                                  └→ push!(PYJLFREEVALUES, idx2)   ← BOOM: flag already set
                                     ConcurrencyViolationError!
```

1. push! needs to grow the vector (exceeds capacity), so it allocates
2. The allocation triggers Julia GC, which runs finalizers
3. A finalizer calls Py_DecRef (because enqueue sees the GIL is held on this same thread), dropping refcount to 0, which triggers _pyjl_dealloc re-entrantly on the same vector

This is especially likely during shutdown (at_jl_exit → jl_atexit_hook), because jl_gc_run_all_finalizers runs finalizers on the calling thread (the main thread, which holds the GIL).  
Many Py objects are finalized at once, repeatedly pushing to PYJLFREEVALUES, and each push that triggers a reallocation can cause the re-entrant chain.

Disabled Julia's GC during the critical vector modifications, so that push!/pop! cannot trigger allocation-based GC,
which prevents the finalizer chain from re-entering.
Added a SpinLock as defense-in-depth against true multi-thread races.

### DIFF
--- a/src/JlWrap/C.jl
+++ b/src/JlWrap/C.jl
@@ -25,6 +25,8 @@ const PyJuliaBase_Type = Ref(C.PyNULL)
 const PYJLVALUES = []
 # unused indices in PYJLVALUES
 const PYJLFREEVALUES = Int[]
+# lock protecting PYJLVALUES and PYJLFREEVALUES from concurrent modification
+const PYJL_LOCK = Threads.SpinLock()
 
 function _pyjl_new(t::C.PyPtr, ::C.PyPtr, ::C.PyPtr)
     alloc = C.PyType_GetSlot(t, C.Py_tp_alloc)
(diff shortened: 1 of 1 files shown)


cjdoris:
Hi, thanks for the PR, I'm glad you isolated the issue you're having.

I think the approach taken for the fix is not the right one, though. Turning Julia's GC off and on like this seems fragile (what happens if another concurrent task also toggles GC, it's not guaranteed that this will end up in the right state at the end). And maybe slow?

I think there is a simpler solution: since the issue just occurs in finalising/deallocating objects, let's avoid resizing `PYJLFREEVALUES` there. We can do this by ensuring that `PYJLFREEVALUES` is always the same length as `PYJLVALUES` (by pushing 0 whenever we push to `PYJLVALUES`, which does not happen in finaliser/deallocator code) and have a `PYJLFREEVALUECOUNT` which tells us how many elements of `PYJLFREEVALUES` we're actually using. This should avoid the re-entrancy issues, since now the part between the spin locks is just a simple array read/write without allocation.

I think the spin locks themselves are a good idea, to avoid concurrent writes to these variables.

dpinol:
> I think the approach taken for the fix is not the right one, though. 

@cjdoris Good point, I applied your feedback. I suggest squashing before merging

cjdoris:
Thanks!

PR #763, opened 2026-04-21
fix: random crash atexit with multithread julia

dpinol:
Invoke Base._atexit instead of jl_atexit_hook to avoid invoking finalizers, which might crash with multithread julia

### DIFF
--- a/pysrc/juliacall/__init__.py
+++ b/pysrc/juliacall/__init__.py
@@ -249,10 +249,15 @@ def init():
     # including running finalizers for any objects that still exist
     @atexit.register
     def at_jl_exit():
-        jl_atexit_hook = lib.jl_atexit_hook
-        jl_atexit_hook.argtypes = [c.c_int]
-        jl_atexit_hook.restype = None
-        jl_atexit_hook(0)
+        jl_eval(b'''
+                try 
+                    Base._atexit(Int32(0))
+                    flush(Base.stdout)
+                    flush(Base.stderr)
+                catch e 
+                    @error "Error during Julia atexit" e
+                end;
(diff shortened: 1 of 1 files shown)


cjdoris:
Hmm I'd rather not rely on any more Julia internals (`Base._atexit`).

Perhaps `jl_atexit_hook` is being called from multiple threads and is not thread safe? Perhaps a lock and ensuring it only gets called once will.fox your issue.

Do you have a MWE for your issue?

dpinol:
> Perhaps `jl_atexit_hook` is being called from multiple threads and is not thread safe? Perhaps a lock and ensuring it only gets called once will.fox your issue.

* I don't see any lock at `jl_atexit_hook`: https://github.com/JuliaLang/julia/blob/master/src/init.c#L256
* Lock is at `_atexit`: https://github.com/JuliaLang/julia/blob/master/base/initdefs.jl#L520

> Do you have a MWE for your issue?

Will work on it

dpinol:
The MWE. Must be done with juliacall 0.9.34.
```bash
export JULIA_NUM_THREADS=6
export PYTHON_JULIACALL_HANDLE_SIGNALS=yes
while pytest pytest/test_all.py -k test_import ; do 
   true
done
```
Without the patch, I stops (curiously it only logs "Segmentation fault" from bash, but not from zsh) after a few loops.
With the patch, it didn't fail after 100 iterations

dpinol:
@cjdoris Could you reproduce the crash with  https://github.com/JuliaPy/PythonCall.jl/pull/763#issuecomment-4486858545 ?

thanks

PR #767, opened 2026-05-16
remove PYNULL_CACHE

cjdoris:


cjdoris:
Comments on the benchmark results:
- Allocations have gone way up. This is expected because we are no longer re-using `Py` objects - they are always recreated, which is an allocation.
- Runtime of things not using `pydel!` are about the same.
- Runtime of things involving `pydel!` are about 1.1x slower.
- So the above allocations are not much of an issue nowadays. They used to be - I suppose Julia is either optimising the allocations away (e.g. putting them on the stack) or the allocator is better these days.
- There's actually very little time difference between the `pydel!` and non-`pydel!` benchmarks, suggesting that `pydel!` is not an optimisation any more (and wasn't much of one before). The main benefit is to ensure that a Python object is freed as early as possible.

Upsides of this PR:
- Code is simpler.
- Removes a race condition on `PYNULL_CACHE` that is causing errors with free-threaded Python.

Downsides:
- Small-Python-object-heavy loops are maybe 10% slower than they could be using `pydel!`.

On balance, this PR is a win.

Thank you to @MilesCranmer for adding the airspeed-velocity stuff earlier - I think this is the first PR where it has proven really beneficial.

PR #768, opened 2026-05-16
Load CondaPkg lazily, and drop Pkg as a dependency (take 2)

JamesWrigley:
Redo of #752 with preferences. Since the CondaPkg change affects module loading I made it only apply when the preference is set rather than the preference or environment variable. Also added a mention to the docs of the priority of preferences vs environment variables.


JamesWrigley:
Closing in favour of #769.

PR #769, opened 2026-05-17
skip importing CondaPkg if not needed (take 3)

cjdoris:
Simpler approach to #768


cjdoris:
@JamesWrigley how about this version? I don't think the getpref stuff needs to be made more complicated just for this one purpose, so this uses `@load_preference` directly.

I also removed the loading of `Pkg` from JuliaCall too - it was only loaded to activate the correct project, but instead I've changed it to use `--project` to start Julia in the correct project.

src/C/C.jl
```diff
@@ -9,10 +9,14 @@ using ..Utils
...
 using Base: @kwdef
 using UnsafePointers: UnsafePtr
-using CondaPkg: CondaPkg
-using Pkg: Pkg
 using Libdl:
     dlpath, dlopen, dlopen_e, dlclose, dlsym, dlsym_e, RTLD_LAZY, RTLD_DEEPBIND, RTLD_GLOBAL
+using Preferences: @load_preference
+
+# do not load CondaPkg if the exe preference is set to something else
+if @load_preference("exe", "") in ("", "@CondaPkg")
```
JamesWrigley: Slightly simpler I think:
```suggestion
if @load_preference("exe", "@CondaPkg") == "@CondaPkg"
```

But overall it looks fine 👍

JamesWrigley:
Thank you very much for getting that over the line :) Out of curiosity, when do you plan the next release?

cjdoris:
I've just merged a flurry of bugfixes so I'll get a release out now.

PR #772, opened 2026-05-18
Fix PyDict REPL completion without GIL

Beforerr:
## Summary

In normal REPL operation, the main Python thread can still hold the GIL while Julia runs completion on an interactive/default thread. Dict-key completion then evaluates a `PyDict` from that completion thread, which is not safe to call Python directly.

This routes the Python-touching pieces through the existing main-thread machinery when needed:

- `keys(::PyDict)` iteration used by `REPL.REPLCompletions.find_dict_matches`
- two-arg `show(::Py)`, because completion calls `repr(key)` and default `PyDict` keys are `Py`

If the current task already owns the GIL, the call stays direct. Otherwise it dispatches to `C.on_main_thread`.

## MWE

This reproduces the crash on `main` here with `julia --project -t 2`, reaching `PyObject_GetIter` from `iterate(::KeySet{<:PyDict})` inside `REPLCompletions.find_dict_matches`.

```julia
using PythonCall, REPL

d = PyDict(Dict("foo" => 1, "bar" => 2))
t = Threads.@spawn REPL.REPLCompletions.completions("d[", 2, @__MODULE__)
wait(t)
fetch(t)
```


cjdoris:
I'm uncertain what problem you're trying to solve here.

In your MWE you unlock the GIL. But in normal operation at the REPL, the GIL will be locked on the main thread, and the issue is that REPL completion is being done from the interactive thread.

So I think your MWE doesn't match the observed problem, if I have understood correctly. And if so, then sprinkling `GIL.@lock` around will not fix the problem, it will deadlock the interactive thread waiting for the GIL to be unlocked, which will not happen (as it is locked by the main thread).

An actual solution to this would be to use the on-main-thread machinery - see `Base.getproperty(::Py)` in `Py.jl`. Possibly overloading `REPL.REPLCompletions.find_dict_matches` would suffice although this is an internal function.

Beforerr:
Thanks, you were right. The original MWE was testing the wrong shape and `GIL.@lock` could deadlock when completion runs off the Python thread while the main thread owns the GIL.

I rewrote this to use the existing main-thread path instead. The only wrapped paths now are `PyDict` key iteration for `find_dict_matches` and `show(::Py)` for the `repr(key)` call.

Checked with focused tests and a `-t 2` spawned completion case.

Beforerr:
Updated the PR body/MWE to match the normal REPL shape

Issue #762, opened 2026-04-21
Improve juliacall startup time?


zengmao:
Currently, it takes more than 2 seconds to load the package `juliacall` in Python, before importing any Julia package or running Julia code:
```
$ time python -c "from juliacall import Main as jl"

real	0m2.355s
user	0m3.104s
sys	0m0.156s
```

This can be a disadvantage compared with calling C++ instead of Julia to offload heavy computations outside Python, depending on the workflow. Is there a way to compile a Julia system image to make the startup much faster?

Issue #600, opened 2025-04-01
Can't use PythonCall in Custom System Image


BenCurran98:
**Affects:** PythonCall

**Describe the bug**
I'm trying to build a Julia system image that uses PythonCall.jl. I can build the system image, but when I try to launch Julia with it it crashes on initialisation. 

As a MWE, I've set up a package [here](https://github.com/BenCurran98/CondaIssue.jl) that just uses PythonCall.jl. You can then run the following code to create a system image with this package:

```julia
using Pkg
Pkg.add(url = "https://github.com/BenCurran98/CondaIssue.jl")
Pkg.add("PackageCompiler.jl")
using PackageCompiler
create_sysimage(["CondaIssue"]; sysimage_path = "CondaSysImage.so")
```

This builds successfully, but when I run it I get the following error:
```
julia -JCondaSysImage.so --project=.
```

```julia
    CondaPkg Found dependencies: /home/ben/.julia/packages/PythonCall/WMWY0/CondaPkg.toml
fatal: error thrown and no exception handler available.
InitError(mod=:C, error=MethodError(f=Base.Cmd, args=(Array{Nothing, 1}(dims=(1,), mem=Memory{Nothing}(1, 0x7183379f3380)[nothing]),), world=0x0000000000006869))
jl_method_error_bare at /cache/build/builder-amdci5-5/julialang/julia-release-1-dot-11/src/gf.c:2254
jl_method_error at /cache/build/builder-amdci5-5/julialang/julia-release-1-dot-11/src/gf.c:2272
jl_lookup_generic_ at /cache/build/builder-amdci5-5/julialang/julia-release-1-dot-11/src/gf.c:3106 [inlined]
ijl_apply_generic at /cache/build/builder-amdci5-5/julialang/julia-release-1-dot-11/src/gf.c:3121
jl_apply at /cache/build/builder-amdci5-5/julialang/julia-release-1-dot-11/src/julia.h:2157 [inlined]
do_call at /cache/build/builder-amdci5-5/julialang/julia-release-1-dot-11/src/interpreter.c:126
eval_value at /cache/build/builder-amdci5-5/julialang/julia-release-1-dot-11/src/interpreter.c:223
eval_stmt_value at /cache/build/builder-amdci5-5/julialang/julia-release-1-dot-11/src/interpreter.c:174 [inlined]
eval_body at /cache/build/builder-amdci5-5/julialang/julia-release-1-dot-11/src/interpreter.c:670
[...]

Issue #436, opened 2023-12-28
Using a sysimage with PythonCall already imported causes error during juliacall initialization


brian-dellabetta:
**Affects:** Both

**Describe the bug**
I would like to use a custom sysimage of our julia package which has PythonCall.jl as a dependency. The image builds fine, and is useable with `julia --sysimage=/path/to/myimage.so`, but when I try to use it as the custom sysimage with juliacall, I hit the following initialization error:
```
PYTHON_JULIACALL_SYSIMAGE=/path/to/myimage.so python
>>> from juliacall import Main as jl
fatal: error thrown and no exception handler available.
InitError(mod=:C, error=AssertionError(msg="CTX.which == :PyCall"))
#35 at /home/xxx/.julia/packages/PythonCall/wXfah/src/cpython/context.jl:144
with_gil at /home/xxx/.julia/packages/PythonCall/wXfah/src/cpython/gil.jl:10 [inlined]
with_gil at /home/xxx/.julia/packages/PythonCall/wXfah/src/cpython/gil.jl:9 [inlined]
init_context at /home/xxx/.julia/packages/PythonCall/wXfah/src/cpython/context.jl:141
__init__ at /home/xxx/.julia/packages/PythonCall/wXfah/src/cpython/CPython.jl:21
jfptr___init___72284 at /home/xxx/.julia/sysimages/myimage.so (unknown line)
_jl_invoke at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-9/src/gf.c:2758 [inlined]
ijl_apply_generic at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-9/src/gf.c:2940
jl_apply at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-9/src/julia.h:1880 [inlined]
jl_module_run_initializer at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-9/src/toplevel.c:75
_finish_julia_init at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-9/src/init.c:855
julia_init at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-9/src/init.c:804
ijl_init_with_image at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-9/src/jlapi.c:66 [inlined]
ijl_init_with_image at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-9/src/jlapi.c:55
[...]

Issue #129, opened 2022-03-07
Trouble when using a sysimage with PythonCall


ericphanson:
I'm having trouble when building a sysimage with PythonCall-- or rather, I can built it OK, but then when I use it, like

```julia
julia --project=myproject --sysimage=mysysimage.so -e "using Pkg; Pkg.instantiate()"
```

I get
```julia
#29 92.79 fatal: error thrown and no exception handler available.
#29 92.79 InitError(mod=:C, error=ErrorException("no environment in the LOAD_PATH depends on CondaPkg"))
#29 92.84 error at ./error.jl:33
#29 92.84 _jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2247 [inlined]
#29 92.84 jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2429
#29 92.84 _resolve_top_env at /root/.julia/packages/CondaPkg/GoPlj/src/resolve.jl:16
#29 92.84 _jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2247 [inlined]
#29 92.84 jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2429
#29 92.84 #resolve#17 at /root/.julia/packages/CondaPkg/GoPlj/src/resolve.jl:222
#29 92.84 _jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2247 [inlined]
#29 92.84 jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2429
#29 92.84 resolve at /root/.julia/packages/CondaPkg/GoPlj/src/resolve.jl:209
#29 92.84 _jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2247 [inlined]
#29 92.84 jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2429
#29 92.84 envdir at /root/.julia/packages/CondaPkg/GoPlj/src/env.jl:66
#29 92.85 init_context at /root/.julia/packages/PythonCall/Z6DIG/src/cpython/context.jl:56
#29 92.86 __init__ at /root/.julia/packages/PythonCall/Z6DIG/src/cpython/CPython.jl:21
#29 92.87 jfptr___init___72494 at /deps.so (unknown line)
#29 92.87 _jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2247 [inlined]
[...]

PR #773, opened 2026-05-19
Support baking PythonCall into a juliacall system image (opt-in)

ncudlenco:
## Summary

Lets `PythonCall` itself be compiled into a juliacall system image (opt-in). With it baked in, the `using PythonCall` that `import juliacall` performs becomes a memory-map instead of a multi-second load+compile. In a containerised juliacall workload we measured, fresh-process startup drops from ~18 s to ~1.9 s. Off by default - no behaviour change unless enabled. It's meant as a low-risk interim until the automatic fix discussed in #436 (or related work) lands; it doesn't conflict with or preclude that.

## Motivation

In short-lived Python processes that embed Julia via juliacall - serverless or autoscaled containers (AWS Lambda, queue workers, CI jobs) that start, handle one request, and exit - there's no long-lived process to amortise Julia start-up. Every cold start pays it again, and the dominant part is the `using PythonCall` that `import juliacall` runs to bring the bridge up.

Measured, fresh container to first call:

|                                              | container init |
|----------------------------------------------|----------------|
| stock juliacall (only app packages baked)    | ~18 s          |
| `PythonCall` also baked into the sysimage    | ~1.9 s         |

`PYTHON_JULIACALL_SYSIMAGE` already exists, but a sysimage with only the application packages still runs `using PythonCall` every cold start. Baking `PythonCall` in too is the fix, but it currently fails: when `PythonCall` is in the sysimage its `__init__` runs during `jl_init_with_image`, before juliacall's bootstrap has set `Main.__PythonCall_libptr`. `init_context()` decides "embedded" only from that global, so it sees "not embedded"; since Python is mid-`import juliacall`, `init_juliacall()` then errors with `'juliacall' module already exists`.

## Change

Add one module-level flag, `PythonCall._is_embedded`, that the user flips inside a PackageCompiler `script=` to mark the resulting sysimage as embedded. At runtime, `PythonCall.__init__` reads the flag and takes the embedded path; libpython is opened from the existing `lib` preference / `JULIA_PYTHONCALL_LIB`. If the sysimage is loaded by a julia process with no running Python (e.g. a `Base.compilecache` child), the embedded path rolls back and PythonCall stays inactive instead of erroring. Docs include a worked PackageCompiler example.

## Backward compatibility

Backwards compatible since the flag is off by default. The classic juliacall path and the non-embedded path are unchanged; `lib` and `exe` preferences keep their existing semantics. No new preferences or environment variables are introduced.

## Testing

The new code path only runs when explicitly enabled, so default behaviour is unchanged and the existing test suites are unaffected.

Verified manually on Linux (AWS Lambda base image `public.ecr.aws/lambda/python:3.12`) and Windows (desktop, Python 3.13). A plain `julia.exe --sysimage=` load of a baked sysimage leaves PythonCall inactive without erroring. Not tested on macOS. No automated test is included because exercising it requires building a system image in CI, which is slow - happy to add one, or document the steps, if you'd prefer.

## Related issues

Relates to #436 and #600 - this unblocks their use case (baking PythonCall into a custom system image) as an interim. The root-cause fix described in #436 - resetting PythonCall's sysimage-persisted state in `__init__` so it works with no opt-in - is the longer-term path; this PR doesn't attempt or preclude it.

Also relevant to #762 ("Improve juliacall startup time?", which explicitly asks about compiling a system image) and a building block for #76 ("Compile and use custom sysimages automatically"). Does not address #129 (a different failure: `no environment in the LOAD_PATH depends on CondaPkg`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 
 ## Unreleased
+* Support baking `PythonCall` into a juliacall system image via the new opt-in
+  `embedded` preference / `JULIA_PYTHONCALL_EMBEDDED` option, removing the
+  `using PythonCall` cost from cold start. No behaviour change unless opted in.
 * Added option `lib` to JuliaCall. Setting this will skip the discovery subprocess.
 * Bug fixes.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,8 @@
 # Changelog
 
 ## Unreleased
-* Support baking `PythonCall` into a juliacall system image via the new opt-in
-  `embedded` preference / `JULIA_PYTHONCALL_EMBEDDED` option, removing the
-  `using PythonCall` cost from cold start. No behaviour change unless opted in.
+* Support baking `PythonCall` into a juliacall system image via
+  `PythonCall._is_embedded[] = true` in a PackageCompiler `script=`.
 * Added option `lib` to JuliaCall. Setting this will skip the discovery subprocess.
 * Bug fixes.
 
--- a/docs/src/juliacall.md
+++ b/docs/src/juliacall.md
@@ -115,6 +115,26 @@ systems that may be readonly. Note that the project set in
 `PYTHON_JULIACALL_PROJECT` *must* already have PythonCall.jl installed and it
 *must* match the JuliaCall version, otherwise loading Julia will fail.
 
+### Baking PythonCall into a system image
+
+For the fastest possible startup you can compile `PythonCall` itself (alongside
+your own packages) into a system image with
+[PackageCompiler.jl](https://github.com/JuliaLang/PackageCompiler.jl), so that
+the `using PythonCall` performed at startup is a memory-map rather than a load.
+
+When `PythonCall` is baked into the system image its `__init__` runs *during*
+`jl_init_with_image`, before juliacall's bootstrap has defined the
+`Main.__PythonCall_libptr` global it normally uses to detect that it is
+embedded. To support this, set the `embedded` preference (or the
+`JULIA_PYTHONCALL_EMBEDDED=yes` environment variable) together with the `lib`
+preference / `JULIA_PYTHONCALL_LIB` pointing at the running interpreter's
+libpython. With `embedded` set, PythonCall takes the embedded path even without
+the global and opens libpython by path (it is already loaded in the process, so
+this is just a handle). The default is `no`, leaving normal behaviour
+unchanged. Use this together with `PYTHON_JULIACALL_SYSIMAGE` (below), and
+`PYTHON_JULIACALL_EXE` / `PYTHON_JULIACALL_PROJECT` so juliacall resolves the
+baked environment directly.
+
 ## [Configuration](@id julia-config)
 
 Some features of the Julia process, such as the optimization level or number of threads, may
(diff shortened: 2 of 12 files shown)


cjdoris:
Looks good. My main feedback is about the docs - could you document the basic steps to actually create a sysimg?

I don't think you need to go much into the why (the current second paragraph) nor mention `Main.__PythonCall_libptr` (which is internal). Instead, you can pretty much just say that you need to set the prefs `embedded=true` and `lib=...`. But what's not totally obvious is how you set up a project and these prefs and use PackageCompiler to actually make the sysimg.

cjdoris:
Another thought - if you know that you are embedded, then you can find the libptr (by calling into the C-API functions, which will be globally available in this case). We only don't do this because it's easier to directly communicate the ptr itself by setting `Main.__PythonCall_libptr` from Python.

Therefore if we can "bake in" the fact that PythonCall is embedded into the sysimg, then we won't need any of these preferences. I wonder if we could simply do `@eval PythonCall _is_embedded=true` or something when we make the sysimg, so it's baked into PythonCall, then test for this variable in `PythonCall.__init__`?

ncudlenco:
> Looks good. My main feedback is about the docs - could you document the basic steps to actually create a sysimg?
> 
> I don't think you need to go much into the why (the current second paragraph) nor mention `Main.__PythonCall_libptr` (which is internal). Instead, you can pretty much just say that you need to set the prefs `embedded=true` and `lib=...`. But what's not totally obvious is how you set up a project and these prefs and use PackageCompiler to actually make the sysimg.

---

> Another thought - if you know that you are embedded, then you can find the libptr (by calling into the C-API functions, which will be globally available in this case). We only don't do this because it's easier to directly communicate the ptr itself by setting `Main.__PythonCall_libptr` from Python.
> 
> Therefore if we can "bake in" the fact that PythonCall is embedded into the sysimg, then we won't need any of these preferences. I wonder if we could simply do `@eval PythonCall _is_embedded=true` or something when we make the sysimg, so it's baked into PythonCall, then test for this variable in `PythonCall.__init__`?

---

Thanks - both points shaped the rewrite. The flag is now baked into the sysimage as you suggested, the docs section has a worked PackageCompiler example, and `Main.__PythonCall_libptr` is no longer mentioned. libpython is opened via the existing `lib` preference (`JULIA_PYTHONCALL_LIB`) - no new preferences or env vars.

I also explored the C-API angle. It works in standalone tests but I couldn't get it to resolve libpython reliably during the sysimage init window across all targets, so the PR sticks with the explicit `lib` preference. Happy to dig deeper if you'd like more detail.

One implementation note: I used `const _is_embedded = Ref(false)` rather than a rebound non-const global, so the C submodule can `import` the name once and read it without parentmodule indirection. User API is the same one line in `script=`. Happy to switch to the literal `@eval` form if you'd prefer.

P.S. Also added you as a Write collaborator on buchi-labortechnik-ag/PythonCall.jl so you can push directly to this branch if you'd like to make changes yourself.

cjdoris:
Is the issue entirely that when in a sysimg, PythonCall is loaded at Julia init, and therefore doesn't see `__PythonCall_libptr`, which is set later?

In which case we could simply pass the libptr via an environment variable instead that we set before we load Julia. This way it gets found whenever PythonCall is loaded, and there is no sysimg-baking required.

This seems like a much more minimal change?

ncudlenco:
> Is the issue entirely that when in a sysimg, PythonCall is loaded at Julia init, and therefore doesn't see `__PythonCall_libptr`, which is set later?
> 
> In which case we could simply pass the libptr via an environment variable instead that we set before we load Julia. This way it gets found whenever PythonCall is loaded, and there is no sysimg-baking required.
> 
> This seems like a much more minimal change?

Yes, this is exactly the issue. Your approach is much cleaner, and works with my setup. I will close this PR in favor of #781. Thanks a lot for the quick turnaround, this unblocks my use case entirely 🙏

Issue #619, opened 2025-05-29
PYTHON_JULIACALL_BINDIR  not effective?


xgdgsc:
**Affects:**  JuliaCall  
From my understanding it still tries to connect the internet when call 
https://github.com/JuliaPy/PythonCall.jl/blob/0a7d49f274147e9876fee47981d9aa5459f6eb8c/pysrc/juliacall/__init__.py#L160C44-L160C54  ? 
**Describe the bug**
PYTHON_JULIACALL_BINDIR  not effective. It still tries to connect the internet and install julia?

![Image](https://github.com/user-attachments/assets/3c678ddd-f1dd-4747-8f9d-e33fb62dc391)

**Your system**
linux centos 7

**Additional context**

xgdgsc:
I don' t understand why doing ` from juliacall import Main as jl` would make it trying things online? Can we just make it totoally offline when importing and only do online stuff with `juliapkg` ?

cjdoris:
Ah the docs maybe aren't clear, if you are trying to use a pre-installed Julia, set `PYTHON_JULIAPKG_EXE`.

xgdgsc:
Thanks for the tip. Is there a way to import without any attempt to do online staff? I want to share a depot to other python user on same machine without any knowledge of julia to use a self hosted git master julia package. What' s the recommended way to do something like this now?

cjdoris:
I've not had to do that before so not sure, but some combination of `PYTHON_JULIAPKG_EXE`, `PYTHON_JULIAPKG_PROJECT`, `PYTHON_JULIAPKG_OFFLINE`, `JULIA_DEPOT` might help.

For a bit on how juliapkg works: it will only download things if it needs to, namely it downloads Julia if you don't have a compatible version already and it downloads packages when dependencies change. So if you resolve once it will not go online again unless any of your dependencies change.

So one way to solve your problem would be to resolve once and then copy the whole filesystem state to your offline system.

Let me know if you figure it out - would be good to add this to the FAQs.

cjdoris:
Thinking a bit more:
- Setting `PYTHON_JULIAPKG_EXE` will use a specific Julia, so won't download one.
- Setting `PYTHON_JULIAPKG_OFFLINE=yes` will prevent any packages from being downloaded.
- However this means the Julia project used will not have any packages. So you should create the project you need yourself and set `PYTHON_JULIAPKG_PROJECT` to its path.
[...]

PR #774, opened 2026-05-19
juliacall: optionally supply libjulia path/bindir to skip the discovery subprocess

ncudlenco:
## Summary

Adds **opt-in** `libpath` / `default_bindir` juliacall options so the path to libjulia and `Sys.BINDIR` can be supplied directly, skipping the short-lived Julia process `init()` otherwise spawns at startup just to discover them. No behaviour change unless both are set.

## Motivation

In **short-lived Python processes that embed Julia via juliacall** (serverless / autoscaled containers), every fresh process pays `import juliacall` startup with no long-lived server to amortise it. During that startup `init()` launches a whole separate Julia process — `julia -O0 --compile=min -e 'print(Libdl.dlpath("libjulia"), Sys.BINDIR)'` — solely to discover two strings. That is an entire extra Julia interpreter boot on the critical path of *every cold start*. In a pre-built container or system image those paths are static and known at build time, so the subprocess is pure repeated waste.

## Change

Two new juliacall options (same `-X` / `PYTHON_JULIACALL_*` mechanism as the rest):

- `libpath` → `PYTHON_JULIACALL_LIBPATH`
- `default_bindir` → `PYTHON_JULIACALL_DEFAULT_BINDIR`

When **both** are set the discovery subprocess is skipped and the supplied values used; otherwise behaviour is exactly as before. `docs/src/juliacall.md` config table and `CHANGELOG.md` updated.

## Backward compatibility

No behaviour change unless both options are set. Values are validated with the existing `path_option(..., check_exists=True)` helper, so a bad path fails fast with the standard error message.

## Testing

The new options only take effect when explicitly set, so default behaviour is unchanged and the existing test suites are unaffected.

Independent of, and complementary to, the system-image PR (#773): together they remove the two biggest fixed costs from a cold juliacall start — loading `PythonCall`, and this extra Julia process.

## Related issues

Relates to #762 ("Improve juliacall startup time?") — this removes a whole extra Julia process from every cold start. Adjacent to #619 (offline / `BINDIR` startup); note #619's specific complaint (juliapkg still attempting a network install) is a different concern this PR does not by itself resolve.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+* JuliaCall: added `libpath` / `default_bindir` options
+  (`PYTHON_JULIACALL_LIBPATH` / `PYTHON_JULIACALL_DEFAULT_BINDIR`) to supply
+  libjulia's path and `Sys.BINDIR` directly, skipping the discovery subprocess
+  at startup. Behaviour is unchanged unless both are set.
+
 ## 0.9.34 (2026-05-18)
 * Bug fixes.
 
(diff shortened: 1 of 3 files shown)


cjdoris:
Closing in favour of #779 (a modified version of this branch). Thanks for the PR!

PR #776, opened 2026-05-19
Reject PyArray buffer with suboffsets

### LINKED ISSUE #775: `PyArray` ignores `suboffsets`
**Affects:** PythonCall

**Describe the bug**
When constructing a `PyArray` the `suboffsets` are ignored. In almost all real world use cases I could find `suboffsets` is NULL and should be ignored, but in theory it should be checked and an error should be thrown saying it isn't supported. `suboffsets` allow having an array stored in PIL-style, but as the example below shows, if treated like a regular strided array we get garbage.

MWE generated by Claude Code:
```julia
using PythonCall

tb = pyimport("_testbuffer")
nd = tb.ndarray(
    pylist([10, 11, 12, 20, 21, 22]),
    shape=pylist([2, 3]),
    format="i",
    flags=tb.ND_PIL | tb.ND_WRITABLE,
)

mv = pybuiltins.memoryview(nd)
println("Python sees:")
println("  shape      = ", mv.shape)
println("  strides    = ", mv.strides)
println("  suboffsets = ", mv.suboffsets)
println("  tolist     = ", mv.tolist())

println()
println("PyArray sees:")
a = PyArray(nd; array=false, buffer=true)
@show typeof(a)
@show size(a)
@show a.strides
@show collect(a)
```

```sh
Python sees:
  shape      = (2, 3)
  strides    = (8, 4)
  suboffsets = (0, -1)
  tolist     = [[10, 11, 12], [20, 21, 22]]

PyArray sees:
typeof(a) = PyArray{Int32, 2, true, false, Int32}
size(a) = (2, 3)
a.strides = (8, 4)
collect(a) = Int32[1081565312 31157 1081565324; 1081565324 31157 10]
```

**Your system**
Julia Version 1.12.6
Commit 15346901f00 (2026-04-09 19:20 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × Intel(R) Core(TM) i7-10700KF CPU @ 3.80GHz
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, skylake)
  GC: Built with stock GC
Threads: 1 default, 1 interactive, 1 GC (on 16 virtual cores)

Python Version 3.14.4
PythonCall Version: current main

nhz2:
Fixes #775 

Generated with help from Claude Code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## Unreleased
+* Bug fixes.
+
 ## 0.9.34 (2026-05-18)
 * Bug fixes.
 
--- a/src/Wrap/PyArray.jl
+++ b/src/Wrap/PyArray.jl
@@ -531,6 +531,8 @@ end
 function PyArraySource_Buffer(x::Py)
     memview = pybuiltins.memoryview(x)
     buf = C.UnsafePtr(C.PyMemoryView_GET_BUFFER(memview))
+    buf.suboffsets[] == C_NULL ||
+        error("PyArray does not support buffers with non-trivial suboffsets (PIL-style indirect layout)")
     PyArraySource_Buffer(x, memview, buf)
 end
 
--- a/test/Wrap.jl
+++ b/test/Wrap.jl
@@ -91,6 +91,17 @@
     end
 end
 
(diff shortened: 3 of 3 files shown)


cjdoris:
Thanks!

PR #779, opened 2026-05-24
774 v2

cjdoris:
Alternative version of #774.

In this version we just add one new option `lib` and re-use the existing `bindir` option.

We have sensible discovery/defaults:
- If `exe` is not set but `bindir` is, then set `exe={bindir}/julia`.
- If `lib` is not set, then launch a julia subprocess (as we did unconditionally before) to find it. Also find `bindir` if not set.
- If `bindir` is not set, then set it to the directory containing `exe`.

Hence setting `exe` (or `bindir`) and `lib` is sufficient to avoid the julia subprocess.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,7 @@
 # Changelog
 
 ## Unreleased
-* JuliaCall: added `libpath` / `default_bindir` options
-  (`PYTHON_JULIACALL_LIBPATH` / `PYTHON_JULIACALL_DEFAULT_BINDIR`) to supply
-  libjulia's path and `Sys.BINDIR` directly, skipping the discovery subprocess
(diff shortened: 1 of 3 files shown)


cjdoris:
@ncudlenco You didn't allow permission for me to modify your PR so i've made this other branch. I reworked the logic - see the PR description above. So now you only need to set `exe` and `lib` to avoid the subprocess, you shouldn't need to set `bindir` at all.

JamesWrigley:
This is quite cool :eyes: Could we get a new release with it?

cjdoris:
Yes I'll try to get one out in the next week.

cjdoris:
FYI I just merged #782 to get the same benefits as this PR automatically when using juliapkg (the default) - it can now find libjulia from juliapkg, which caches it, so it doesn't require this discovery step.

PR #780, opened 2026-05-30
load python symbols globally when embedded rather than using lib_ptr (which is now NULL in this case)

cjdoris:
Supports #773.


cjdoris:
Not needed if #781 solves #773.

cjdoris:
Not needed.

PR #781, opened 2026-05-31
Libptr by env

cjdoris:
Pass the libptr in juliacall from Python to Julia with an env var instead of by setting a secret variable in Main.

We do this because we can set the env var earlier, in particular, so that PythonCall sees it even if it is loaded during jl_init(), which happens when PythonCall is baked into a system image.

I think this resolves the issue in #773.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Changelog
 
 ## Unreleased
-* Added option `lib` to JuliaCall. Setting this will skip the discovery subprocess.
+* Add option `lib` to JuliaCall. Setting this will skip the discovery subprocess.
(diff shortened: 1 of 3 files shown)


cjdoris:
@ncudlenco this PR changes how libptr is passed into julia from python, so it should now be picked up if PythonCall is inited during jl_init, which is the case when it is baked into a sysimage. Could you verify if this change is sufficient to use such a sysimage? I'm hopeful that this is an alternative to #773 that doesn't require any sysimage-specific handling.

ncudlenco:
I can confirm that this works fine as an alternative for #733. Way more elegant than my attempts 🚀

PR #806, opened 2026-08-14
avoid precompile cache miss from check-bounds option

### LINKED ISSUE #92: If using resolve and with `PYTHON_JULIACALL_CHECK_BOUNDS: yes` will always precompile twice.
MWE:
```
from os import environ
environ["PYTHON_JULIACALL_CHECK_BOUNDS"] = 'yes'

import juliapkg
juliapkg.resolve(force=True)  # precompiles here
import juliacall

jl = juliacall.newmodule("MyPythonProject")
jl.seval("using LocalJuliaPackage")  # precompiles again here
```

`juliapkg.resolve` always runs [precompile](https://github.com/JuliaPy/pyjuliapkg/blob/81dd6a044b7dbdd5bff298cc962cdc8bd60c90f7/src/juliapkg/deps.py#L574C38-L574C40).
It always runs with the [default (auto) setting settings for `--check-bounds`.](https://github.com/JuliaPy/pyjuliapkg/blob/main/src/juliapkg/deps.py#L630-L632).
This creates a compile cache with the `check_bounds=0` flag set.

Loading a package using `juliacall` respects the `PYTHON_JULIACALL_CHECK_BOUNDS` setting.
This thus goes looking for a precompiled copy of the package with
the `check_bounds=1` flag set.
This fails. It thus precompiles everything again.

This happens a lot if the package in question is being concurrently developed in a monorepo along side the python package.

### Closer to real case
<details>
What is really going on is I have a mono-repo.
In it is a python project `MyPythonProject`
a local julia package `LocalJuliaPackage`.
And I want to use that julia package from the python package.
So I haved a `juliapkg.json` store the mono repo,
that remembers to `dev` the local julia package.

Now calling `import juliacall`  in `MyPythonPackage` might fail,
if it on a computer that has never had this installed before
(E.g. CI).
Or if the dependencies of `LocalJuliaPackage` have changed.
So what i do is try to import it.
If it throws and exception, do a `juliapkg.resolve(force=True)` and try again.
This will fix anything that might go wrong that can be fixed that way, and anything else will throw an exception again.

Then after that I then do the normal `juliacall` way of loading things.

```
from os import environ
from MyPythonProject import REPO_ROOT

local_julia_env_dir = REPO_ROOT / "julia"
environ["PYTHON_JULIAPKG_PROJECT"] = str(local_julia_env_dir)


# juliacall must only be imported after setting above enviroment variables
try:
    import juliacall
except Exception:
    # Make sure the issue isn't caused by not having dependencies resolved after LocalJuliaPackage dependency changes
    import juliapkg

    juliapkg.resolve(force=True)
    # now reimport it: either it works and all is well, or it will throw an error again
    import juliacall



jl = juliacall.newmodule("MyPythonProject")
jl.seval("using LocalJuliaPackage")
local_julia_package = jl.LocalJuliaPkg  # reexpose this for to be accessed as `MyPythonProject.julia.local_julia_package
```
In CI I set the `PYTHON_JULIACALL_CHECK_BOUNDS: yes`
since that is a good idea when testing.

</details>

# Possible fixes:

1. Make PyJuliaPkg either respect `PYTHON_JULIACALL_CHECK_BOUNDS`, or provide its own option for that
2. Remove the precompile from within the resolve, or provide an option for that. (Potentially respecting `JULIA_PKG_PRECOMPILE_AUTO`)

Related to: https://github.com/JuliaPy/pyjuliapkg/issues/24

cjdoris:
Fixes https://github.com/JuliaPy/pyjuliapkg/issues/92

Does so by passing precompile-relevant julia args into `juliapkg.resolve()` so that it precompiles with the same args as juliacall uses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## Unreleased
+* Bug fix: avoid precompilation cache miss when `check_bounds` option set.
+
 ## 0.9.35 (2026-06-08)
 * Add option `lib` to JuliaCall. Setting this will skip the discovery subprocess.
 * Add support for using a system image in `juliacall` that has `PythonCall` baked in.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ classifiers = [
     "Operating System :: OS Independent",
 ]
 requires-python = ">=3.10, <4"
-dependencies = ["juliapkg >=0.1.24, <0.2"]
+dependencies = ["juliapkg >=0.1.26, <0.2"]
 
 [dependency-groups]
 dev = [
--- a/pysrc/juliacall/__init__.py
+++ b/pysrc/juliacall/__init__.py
@@ -198,6 +198,18 @@ def init():
         # `exepath` and `project` are set by the user.
         import juliapkg
 
+        # Explicitly resolve so we can pass through julia_args. This ensures that any
+        # precompilation done by juliapkg matches how we launch julia, avoiding
+        # cache misses.
+        julia_args = [
+            "--" + opt[4:].replace("_", "-") + "=" + CONFIG[opt]
+            for opt in [
+                "opt_check_bounds",
+            ]
+            if CONFIG[opt] is not None
+        ]
+        juliapkg.resolve(julia_args=julia_args)
+
         # Find the Julia executable and project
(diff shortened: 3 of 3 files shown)


cjdoris:
Test failures are unrelated (and mysterious).

PR #808, opened 2026-08-21
Add __test__ attribute to AnyValue class for pytest

oxinabox:
I don't entirely understand why this happens.
I think it is some interaction between pytest and the python debugger and it not undestanding python code when that code is not in a `.py` file.

But running juliacall in the python debugger, excent with it set to only break on User Uncaught Exceptions  always breaks like 2-5 times in this code, because it tried to check if it has a `__test__` attribute. Which it does not.

Easiest fix is just to add it.
Though it isn't the most satisfying.

### DIFF
--- a/src/JlWrap/any.jl
+++ b/src/JlWrap/any.jl
@@ -233,6 +233,7 @@ function init_any()
 $("\n"^(@__LINE__()-1))
 class AnyValue(ValueBase):
     __slots__ = ()
+    __test__ = False  # make pytest not try looking for tests in here
     def __repr__(self):
         if self._jl_isnull():
             return "<jl NULL>"


cjdoris:
The fix feels very much like a hack. Can you figure out what the actual issue is?

oxinabox:
> The fix feels very much like a hack. Can you figure out what the actual issue is?

I have spend several afternoons trying to work it out.
And it is beyond me.

I think it is a complicated intraction between exactly how the python debugger in vs-code interacts with the pytest runner.
And I think it is that duriing `pytest --collect` the debugger settings are always "break on any exception" even if the user as said to break only on uncaught exceptions in vs-code.

cjdoris:
Do you have a reproducer?

oxinabox:
I can try and put together a whole pixi enviroment to reproduce and upload a zip file or something.

PR #811, opened 2026-09-02
protect GIL with Julia reentrant mutex

kleinschmidt:
This is an alternative to/elaboration on #810, which uses a ReentrantLock to
prevent `GIL.@lock` and friends from stealing the GIL from other Julia tasks
running on other threads.  The behavior of `PyGILState_Ensure()` is to
immediately change the attached thread state, and it does not block until the
GIL is not locked.  This means that _some_ additional protection is necessary to
prevent multiple julia tasks on different threads from stealing the GIL from
each other leading to UB.

Like #810, it makes the task _sticky_ while the GIL is held.  There isn't really
any sense in which the GIL can be held safely on a non-sticky task, and by
unifying locking the GIL with pinning the locking task to its current thread,
this makes multithreaded code with yield points much safer.

I'm not sure whether we may _also_ want an additional "unsafe" mode without the
reentrant lock/task pinning, although at that point you might as well do the
`PyGILState_` calls directly...

I did not change the behavior of the GC, which is only controlled by the
`PyGILState_Check()` function and doesn't manipulate the GIL itself.  I believe
it's still reasonable, and to the extent that it requires some care to use
correctly, it's not affected by these changes.  However, the GC hook test is
failing and I'm not sure why; when I run it interactively it will _eventually_
run the finalizer and empty the queue; it just doesn't happen on the timeline
that the test expects it.  I've tried adding a yield/timedwait to no avail, so
open for suggestions there.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "PythonCall"
 uuid = "6099a3de-0909-46bc-b1f4-468b9a2dfc0d"
 authors = ["Christopher Doris <github.com/cjdoris>"]
-version = "0.9.35"
+version = "0.9.36"
 
 [deps]
 CondaPkg = "992eb4ea-22a4-4c89-a5bb-47a3300528ab"
--- a/src/GIL/GIL.jl
+++ b/src/GIL/GIL.jl
@@ -25,6 +25,12 @@ if Base.VERSION ≥ v"1.11"
     )
 end
 
+const LOCK = ReentrantLock()
+
+function __init__()
+    Base.lock(LOCK)
+    return nothing
+end
 
 """
     lock(f)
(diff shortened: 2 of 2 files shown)


kleinschmidt:
alright seems like this may be a non-starter given the GIL tests _explicitly_ test for calling into python concurrently 😞

cjdoris:
I think I'm ok with changing the semantics of these macros given they are more-or-less useless currently.

Over on #812 I'm doing something similar but for a different end-goal: handling the GIL automatically so you don't need these macros at all. The final implementation will be fiddly and quite a bit more complicated than this PR, but once it's ironed out I'll back-port the logic to these macros.

PR #814, opened 2026-09-09
Set JULIA_PYTHONCALL_EXE instead of JULIA_PYTHONCALL_EXECUTABLE in juliacall

MilesCranmerBot:
Fixes #813.

JuliaCall set the environment variable `JULIA_PYTHONCALL_EXECUTABLE`, which nothing reads, instead of `JULIA_PYTHONCALL_EXE`. Since PythonCall reads `JULIA_PYTHONCALL_EXE` (with a Preferences fallback) to locate the Python executable, child Julia processes spawned from an embedded Julia did not see the running Python and instead resolved a fresh CondaPkg environment.

This sets the correct variable at the process level before `jl_init`, so it is visible both to PythonCall during initialization (e.g. when baked into a sysimage) and to any child Julia processes.

PR #815, opened 2026-09-09
Add support for `@ENV` as an exe preference

JamesWrigley:
This will force PythonCall to read the executable from the environment variable. Quoting the new docs:
>  The advantage of setting `@ENV` as a preference is
>   that CondaPkg.jl will not be loaded, which saves on load time, while you still
>   keep the ability to change Python environments based on environment
>   variables. This can be helpful when using shared Julia environments on a
>   cluster so that users can easily switch Python environments without modifying
>   the depot preferences.

When implementing this I also noticed that in embedded mode we set `CTX.exe_path` by the preference, which was wrong before (coincidentally fixed just now by #814 :stuck_out_tongue:) and seems a bit fragile, so in 777fac8 I changed it to read it from `sys.executable` instead. Also wrapped a few functions from the stable ABI to help.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -53,12 +53,15 @@ jobs:
         with:
           python-version: ${{ matrix.pyversion }}
       
+      # Always build PyCall (and hence Conda) so that Conda's root env directory exists;
+      # otherwise Conda fails to precompile on cached runners, which can cause errors.
+      # Use a temporary environment because `--project=test` has no manifest on Julia
+      # versions without workspace support (< 1.12).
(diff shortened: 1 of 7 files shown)


JamesWrigley:
AFAICT the failures are unrelated :eyes:

JamesWrigley:
Debugged the CondaPkg test failures to an invalid cached build of Conda.jl being used, fixed in 44aa949 by always building PyCall. The benchmark job failure seems to be because this PR is from my fork and doesn't have privs to comment.

JamesWrigley:
(rebased to fix merge conflicts)

PR #817, opened 2026-09-10
Fix JuliaCall binary-directory discovery for symlinks and wrappers

### LINKED ISSUE #816: JuliaCall 0.9.35 regression: initialization fails when Julia executable is a symlink
**Affects:** JuliaCall

### Describe the bug

JuliaCall 0.9.35 fails during initialization when Julia is discovered through a symlink outside its installation directory. It searches for `sys.so` relative to the symlink's directory rather than Julia's actual binary directory.

The same setup works with JuliaCall 0.9.34. Version 0.9.35 also works when the real executable directory is used instead. The failure can be reproduced with a standalone `import juliacall`.

### Reproduction

Prerequisites: `uv` and an existing Julia 1.12.7 installation. Set `JULIA_REAL` to its actual executable, not a symlink or Juliaup launcher. Run without JuliaCall/JuliaPkg configuration overrides or `juliaup` on the probe's `PATH`.

```bash
JULIA_REAL=/absolute/path/to/julia-1.12.7/bin/julia

work="$(mktemp -d)"
mkdir -p "$work/link/bin"
ln -s "$JULIA_REAL" "$work/link/bin/julia"

uv venv --python 3.14.5 "$work/venv"
uv pip install --python "$work/venv/bin/python" \
  juliacall==0.9.35 juliapkg==0.1.26

PATH="$work/link/bin:/usr/bin:/bin" \
JULIA_DEPOT_PATH="$work/depot" \
"$work/venv/bin/python" -I - <<'PY'
import juliapkg

print("Selected executable:", juliapkg.executable(), flush=True)

import juliacall
PY
```

JuliaPkg selects `<work>/link/bin/julia`. Julia package setup succeeds, but embedded initialization exits with:

```text
ERROR: could not load library "<work>/link/bin/../lib/julia/sys.so"
<work>/link/bin/../lib/julia/sys.so: cannot open shared object file: No such file or directory
```

**Expected:** importing JuliaCall should work through the symlink, as it does through the real executable path.

### Controlled comparison

Each case used a separate Python environment and Julia project, with the same Python, Julia, and JuliaPkg versions:

| JuliaCall | Discovered executable | Result |
|---|---|---|
| 0.9.35 | Symlink | Initialization fails; exit 1 |
| 0.9.35 | Real path | Imports successfully; evaluates `1 + 1` to `2` |
| 0.9.34 | Symlink | Imports successfully; evaluates `1 + 1` to `2` |

### Your system

- Standalone reproduction: Arch Linux, x86_64
- Original failure: GitHub Actions `ubuntu-24.04`, where `/usr/bin/julia` points to `/usr/local/julia1.12.7/bin/julia`
- Python: 3.14.5 in the standalone reproduction
- Julia: 1.12.7

Installed Python packages in the failing environment:

```text
filelock   3.32.6
juliacall  0.9.35
juliapkg   0.1.26
semver     3.0.4
tomli      2.4.1
tomlkit    0.15.1
```

`juliapkg.status()` output, with temporary paths shortened:

```text
JuliaPkg Status
<work>/venv/julia_env/pyjuliapkg/juliapkg.json (empty project)
Julia 1.12.7 @ <work>/link/bin/julia
```

### Additional context

This appears to reintroduce the failure described in [#196](https://github.com/JuliaPy/PythonCall.jl/issues/196), which was fixed in 2022.

[PR #782](https://github.com/JuliaPy/PythonCall.jl/pull/782), included in 0.9.35, supplies `libpath` through `juliapkg.libjulia()`. This skips the subprocess that also discovers `Sys.BINDIR`. With no explicit `bindir`, initialization then falls back to:

```python
bindir = os.path.dirname(exepath)
```

For a symlink such as `/usr/bin/julia`, that produces the wrong directory. Restoring discovery of Julia's actual `Sys.BINDIR`, or otherwise handling the symlink correctly, should address this initialization failure.

schmidma:
Fixes #816.

JuliaCall can locate `libjulia` through JuliaPkg while still deriving the wrong binary directory from the executable path. When that executable is a symlink or wrapper outside the Julia installation, initialization searches for `sys.so` in the wrong location.

Query Julia for `Sys.BINDIR` whenever the binary directory is unknown, preserving any library path or binary directory already supplied. Remove the executable-directory fallback. Using Julia's own directory information also handles wrappers, which resolving symlinks alone would not fix.

This restores a discovery subprocess for default initialization. Configurations where both `libpath` and `bindir` are known still skip that subprocess.

Add subprocess regression tests for a symlink and a POSIX executable wrapper outside the installation directory. Both exercise normal JuliaPkg resolution with its cached library path.

## Validation

The new symlink and wrapper regression tests reproduce the missing `sys.so` failure on unmodified `main` and pass with this patch.

Local Julia-suite validation encounters a GTK3 test failure because GTK is available where the test expects an import exception. The identical failure was reproduced on unmodified `main`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## Unreleased
+* Bug fix: JuliaCall initialization when the Julia executable is a symlink or
+  wrapper outside its installation directory.
 * Bug fix: avoid precompilation cache miss when `check_bounds` option set.
 * Bug fix: `juliacall` set the environment variable `JULIA_PYTHONCALL_EXECUTABLE`
   instead of `JULIA_PYTHONCALL_EXE`, so child Julia processes resolved a new
--- a/pysrc/juliacall/__init__.py
+++ b/pysrc/juliacall/__init__.py
@@ -220,16 +220,15 @@ def init():
     if (libpath is not None) and (exepath is None):
         raise Exception("PYTHON_JULIACALL_EXE is required if PYTHON_JULIACALL_LIB is set.")
 
-    # Find the Julia library, if not specified.
-    if libpath is None:
+    # Discover the Julia library and binary directory if either is missing.
+    if libpath is None or bindir is None:
         cmd = [exepath, '--project='+project, '--startup-file=no', '-O0', '--compile=min',
                '-e', 'import Libdl; print(abspath(Libdl.dlpath("libjulia")), "\\0", Sys.BINDIR)']
-        libpath, found_bindir = subprocess.run(cmd, check=True, capture_output=True, encoding='utf8').stdout.split('\0')
-        CONFIG['libpath'] = libpath
+        found_libpath, found_bindir = subprocess.run(cmd, check=True, capture_output=True, encoding='utf8').stdout.split('\0')
+        if libpath is None:
+            CONFIG['libpath'] = libpath = found_libpath
         if bindir is None:
             CONFIG['bindir'] = bindir = found_bindir
-    if bindir is None:
-        bindir = os.path.dirname(exepath)
     assert os.path.exists(libpath)
     assert os.path.exists(bindir)
 
--- a/pytest/test_issue_816.py
+++ b/pytest/test_issue_816.py
@@ -1,0 +1,89 @@
+import os
+from pathlib import Path
+import subprocess
+import sys
+
+import pytest
+
+
+_CHILD = r"""
+import os
+from pathlib import Path
+import shlex
+import stat
+
+import juliapkg
+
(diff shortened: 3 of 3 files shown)


cjdoris:
Thanks. The PR makes sense but I restructured things like this in the previous version in order to avoid needing to do this init-time discovery in common cases. We did this by having juliapkg pass through the libjulia path via `juliapkg.libjulia()` (and juliapkg caches this at install-time). I'd like to add `juliapkg.bindir()` too so we can retain the faster no-discovery path.

cjdoris:
I made the changes I suggested. Thanks for the PR!

